PR #5, opened 2016-08-28
Add support for enumerated values to field definitions.

brandonedens:
Freescale and Atmel both utilize these fields. Let me know if the implementation is sound. Note the Freescale stuff is encoding its values with #0101x01 which we support. Discussion of that is here:
https://github.com/posborne/cmsis-svd/issues/14

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -161,7 +161,7 @@ impl Register {
     }
 }
 
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
 pub enum Access {
     ReadOnly,
     ReadWrite,
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -305,25 +305,9 @@ impl EnumeratedValues {
 }
 
 #[derive(Debug)]
-pub struct Choice {
-    value: Option<u32>,
-    is_default: Option<bool>,
-}
-
-impl Choice {
-    fn parse(tree: &Element) -> Choice {
-        Choice {
-            value: try!(tree.get_child_text("value")).parse().ok(),
-            is_default: try!(tree.get_child_text("isDefault")).parse().ok(),
-        }
-    }
-}
-
-#[derive(Debug)]
 pub struct EnumeratedValue {
     pub name: String,
     pub description: Option<String>,
-    pub choice: Option<Choice>,
     pub value: u32,
 }
 
@@ -336,7 +320,6 @@ impl EnumeratedValue {
         Some(EnumeratedValue {
             name: try!(tree.get_child_text("name")),
             description: tree.get_child_text("description"),
-            choice: tree.get_child("choice").map(Choice::parse),
             value: try!(parse::u32(try!(tree.get_child("value")))),
         })
     }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -190,6 +190,7 @@ pub struct Field {
     pub name: String,
     pub description: Option<String>,
     pub bit_range: BitRange,
+    pub access: Option<Access>,
     pub enumerated_values: Option<EnumeratedValues>,
 }
 
@@ -201,6 +202,7 @@ impl Field {
             name: try!(tree.get_child_text("name")),
             description: tree.get_child_text("description"),
             bit_range: BitRange::parse(tree),
+            access: tree.get_child("access").map(Access::parse),
 
             enumerated_values:
                 tree.get_child("enumeratedValues").map(EnumeratedValues::parse),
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -255,3 +259,85 @@ impl Defaults {
...
+            is_default: try!(tree.get_child_text("isDefault")).parse().ok(),
+        }
+    }
+}
+
+#[derive(Debug)]
+pub struct EnumeratedValue {
+    pub name: String,
+    pub description: Option<String>,
+    pub choice: Option<Choice>,
```
japaric: I see 'choice' doesn't appear in the cmsis-svd repository. I would simply not parse it if it doesn't appear in the real world.

src/lib.rs
```diff
@@ -255,3 +259,85 @@ impl Defaults {
...
+#[derive(Debug)]
+pub struct Choice {
+    value: Option<u32>,
+    is_default: Option<bool>,
+}
+
+impl Choice {
+    fn parse(tree: &Element) -> Choice {
+        Choice {
+            value: try!(tree.get_child_text("value")).parse().ok(),
```
japaric: This looks wrong. 'value: Option<u32>' indicates that the field may or may not appear but 'try!(tree.get_child_text("value"))' states that the field *must always* appear (because our try! is equivalent to unwrapping). This should be 'tree.get_child_text("value").map(|s| try!(s.parse()))'.

src/lib.rs
```diff
@@ -255,3 +259,85 @@ impl Defaults {
...
+            is_default: try!(tree.get_child_text("isDefault")).parse().ok(),
```
japaric: same as 'value' above

japaric:
Thanks for the PR, @brandonedens.

Overall looks good to me, except for the Choice struct.

The part about the X being a don't care bit and getting parsed as 0 seems reasonable to me. If later we find out that we should preserve the information about the don't care bits then we can change the representation to something like `Value { value: u32, dont_care_mask: Option<u32> }` where `dont_care_mask` would be `0b0001010` if the string is e.g. `101x0x1` (value would be `0b1010001` like today).

brandonedens:
I'm fixing things up according to your feedback. About choice and optional value; didn't have an example and didn't build one. I gathered this stuff from:
http://www.keil.com/pack/doc/CMSIS/SVD/html/group__svd__xml__enum__gr.html
specifically:
<enumeratedValue>
    <name>identifierType</name>
    <description>xs:string</description>
    <choice>
        <value>scaledNonNegativeInteger</value>
        <isDefault>xs:boolean</isDefault>
    </choice>
</enumeratedValue>

and the instances we find in the wild look like:
<enumeratedValue>
<!-- ... ->
    <value>1</value>
</enumeratedValue>

but I agree we should probably just work at what the vendors have given us whether or not its spec. I also wonder if Keil's CMSIS-SVD 1.3.1 is an older/newer specification.

japaric:
@homunkulus try

homunkulus:
:hourglass: Trying commit 9520812 with merge 55e8782...

homunkulus:
:broken_heart: Test failed - [travis](https://travis-ci.org/japaric/svd/builds/156689524)

japaric:
There's a problem with NXP/LPC176x5x_v0.2.svd; there's an EnumeratedValue tag with no Value tag inside: (line 199)

``` xml
<enumeratedValue>
  <name>RESERVED</name>
  <description>Reserved.</description>
  <isDefault>true</isDefault>

</enumeratedValue>
```

brandonedens:
@homunkulus try

japaric:
@homunkulus try

japaric:
@homunkulus try (homunkulus was sleeping :-)

homunkulus:
:hourglass: Trying commit 19f2084 with merge 45ae2ec...

brandonedens:
Most tests pass now more so than before except for:
SIM3L1x8_SVD
which is due to
```html
<description> </description>
```
which causes a panic when 
```rust
c.text.clone()
```
is invoked on line 44. Unsure how to fix that.

japaric:
```
<description> </description>
```

An empty description tag sounds/looks wrong to me. We could consider it as `descrption = None`, or simply blacklist this particular SVD (there's a `blacklist` array in the `generate-tests.sh` script). I'm inclined to blacklist it.

brandonedens:
@homunkulus try

japaric:
*wakes up homu*

@homunkulus try

homunkulus:
:hourglass: Trying commit 147a483 with merge 4793939...

homunkulus:
:sunny: Test successful - [travis](https://travis-ci.org/japaric/svd/builds/157518833)

japaric:
Thank you, @brandonedens!

PR #14, opened 2016-12-04
Basic support for register arrays

istankovic:
Not really sure about the design here, probably it would be better to have a distinct `RegisterArray` type
or something so that the users (e.g. svd2rust) can decide how to handle those themselves. But this 
is a starting point at least.

Another thing to discuss/improve here is the representation of register arrays. Currently it's all exploded, which is not nice. If we're going to create proper arrays on the Rust side, we need to see what to do
with non-integer indices (i.e. the dimIndex case).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,7 +6,7 @@ keywords = ["CMSIS", "SVD", "parser"]
 license = "MIT OR Apache-2.0"
 name = "svd-parser"
 repository = "https://github.com/japaric/svd"
-version = "0.1.2"
+version = "0.1.3"
 
 [dependencies]
 xmltree = "0.3.2"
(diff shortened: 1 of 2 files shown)


japaric:
> Not really sure about the design here, probably it would be better to have a distinct RegisterArray type
or something so that the users (e.g. svd2rust) can decide how to handle those themselves. 

I'd prefer the latter. Basically turn the current `Register` into an enum:

``` rust
enum Register {
    Single { .. }
    Array { .. }
    // ..
}
```

and leave downstream users handle that as they see fit.

@istankovic, could you link me  to a SVD file that uses this "register array" feature?

istankovic:
Yeah, the enum approach is nice and is one of the ideas I had in mind.
I'm going to rework the current implementation.

For an example of an SVD file you can look at
https://github.com/posborne/cmsis-svd/blob/master/data/Atmel/ATSAMD21E17A.svd.

japaric:
Closing in favor of #16

PR #26, opened 2017-04-28
Added error checking

Emilgardis:
This is a pretty large change, and should be carefully looked over.

This commit should be seen as a prototype, as there are some things that can be made more ergonomic and prettier.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -26,46 +26,86 @@
 
 extern crate xmltree;
 
-use std::ops::Deref;
+#[macro_use]
+extern crate error_chain;
+
+use std::ops::{Deref, DerefMut};
+use std::fmt::Debug;
+
+pub mod errors;
 
 use xmltree::Element;
+use errors::{Result, ResultExt};
 
-macro_rules! try {
-    ($e:expr) => {
-        $e.expect(concat!(file!(), ":", line!(), " ", stringify!($e)))
+macro_rules! bail_if_none {
+    ($fun:expr, $error:expr) => {
+        match $fun {
+            ::std::option::Option::Some(val) => val,
+            ::std::option::Option::None => bail!($error)
+        }
+    };
+    ($fun:expr, $error:expr,) => {
+        match $fun {
+            ::std::option::Option::Some(val) => val,
+            ::std::option::Option::None => bail!($error)
+        }
+    };
+    ($fun:expr, $error:expr, $($arg:tt)+) => {
+        match $fun {
+            ::std::option::Option::Some(val) => val,
+            ::std::option::Option::None => bail!($error,$($arg)+)
+        }
     }
 }
 
 mod parse;
 
 /// Parses the contents of a SVD file (XML)
-pub fn parse(xml: &str) -> Device {
-    Device::parse(xml)
+pub fn parse(xml: &str) -> Result<Device> {
+    Device::parse(xml).chain_err(|| "Failed to parse svd file")
 }
 
 trait ElementExt {
-    fn get_child_text<K>(&self, k: K) -> Option<String>
+    fn get_child_text_try<K>(&self, k: K) -> Result<Option<String>>
     where
-        String: PartialEq<K>;
+        String: PartialEq<K>,
+        K: Debug + Clone;
+    fn get_child_try<K>(&self, k: K) -> Result<&Element>
+    where
+        String: PartialEq<K>,
+        K: Debug + Clone;
     fn debug(&self);
 }
 
 impl ElementExt for Element {
-    fn get_child_text<K>(&self, k: K) -> Option<String>
+    fn get_child_text_try<K>(&self, k: K) -> Result<Option<String>>
     where
         String: PartialEq<K>,
+        K: Debug + Clone,
     {
-        self.get_child(k).map(|c| try!(c.text.clone()))
+        match self.get_child(k.clone()) {
+            None => Ok(None),
+            Some(val) => Ok(Some(bail_if_none!(val.text.clone(), "Couldn't get `{:?}`", k),),),
+        }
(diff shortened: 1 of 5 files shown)


Emilgardis:
Derived from support seems to not work as it should, I'll either remove it or fix it asap

homunkulus:
:umbrella: The latest upstream changes (presumably 5ee1e0a) made this pull request unmergeable. Please resolve the merge conflicts.

Emilgardis:
I've removed the derivedfFrom support on registers for now. They weren't properly implemented

src/lib.rs
```diff
@@ -153,12 +227,28 @@ pub struct Interrupt {
...
-    fn parse(tree: &Element) -> Interrupt {
-        Interrupt {
-            name: try!(tree.get_child_text("name")),
-            description: tree.get_child_text("description"),
-            value: try!(parse::u32(try!(tree.get_child("value")))),
-        }
+    fn parse(tree: &Element) -> Result<Interrupt> {
+        let name = bail_if_none!(
+            tree.get_child_text_try("name")?,
+            "Couldn't find name of `interrupt`",
```
japaric: Actually this could be handled in a layered manner:

```
error: <name> not found
caused by: in interrupt #5
caused by: in peripheral GPIO
```

japaric: Never mind all these comments. I had not looked at the produced error messages just yet.

src/lib.rs
```diff
@@ -26,47 +26,87 @@
...
+
+pub mod errors;
 
 use xmltree::Element;
+use errors::{Result, ResultExt};
 
-macro_rules! try {
-    ($e:expr) => {
-        $e.expect(concat!(file!(), ":", line!(), " ", stringify!($e)))
+macro_rules! bail_if_none {
```
Emilgardis: I agree, will change when I have time as I'm currently abroad.

src/lib.rs
```diff
@@ -82,19 +122,32 @@ impl Device {
...
-        Device {
-            name: try!(tree.get_child_text("name")),
-            peripherals: try!(tree.get_child("peripherals"))
-                .children
-                .iter()
-                .map(Peripheral::parse)
-                .collect(),
-            defaults: Defaults::parse(tree),
-            _extensible: (),
+        let peripherals_inter = bail_if_none!(
```
Emilgardis: "inter" here stands for intermediate.

Emilgardis:
@japaric my vision for this was something like you propose here. Wording errors with error-chain has always been something hard for me as the flow of information can be different depending on what error you want to give out.


I think the best approach is to introduce multiple custom erros via:

```
error_chain! {
    errors {
        FieldError(field: String) {
             display("error regarding {}",field)
        }
    }
}
```
to make this whole process easier.

If you could make those SVD files it would certainly be helpful.

With your example would we omit the `caused by` text that is usually used with error-chain?

src/lib.rs
```diff
@@ -113,35 +166,56 @@ pub struct Peripheral {
...
-            registers: tree.get_child("registers")
-                .map(
-                    |rs| {
-                        rs.children
-                            .iter()
-                            .filter_map(Register::parse)
-                            .collect()
+    fn parse(tree: &Element) -> Result<Peripheral> {
+        if tree.name != "peripheral" {
+            bail!("peripheral not peripheral")
```
Emilgardis: I was wondering about this one, it checks if an element of `<peripherals>` is a `<peripheral>`, maybe it should be reworded to be `"Element is <{tree.name}>, expected <peripheral>"` and chained together with `"While parsing peripheral #x of <peripherals>"`

japaric: Yes, that sounds better!

src/lib.rs
japaric: OK. I see.

src/lib.rs
japaric: Thanks!

japaric:
> Wording errors with error-chain has always been something hard for me as the flow of information

I understand. We could probably rework how the parsing is done to provide more precise information and have more control over the granularity of error messages (e.g. peripheral -> register -> filed -> attribute) but I think we should not do that in this PR; we should simply add error message to what we have right now.

> If you could make those SVD files it would certainly be helpful.

Alright. I'll produce some and gist them here.

> With your example would we omit the caused by text that is usually used with error-chain?

Yeah, I think so. This is probably only be used by svd2rust at this time so we should optimize the error mesasges here to make svd2rust errors are clear as possible. If we get more svd users in the future then we can reconsider the wording of the error messages in this repo.

japaric:
@Emilgardis Here are [some SVD files](https://gist.github.com/japaric/d4685881df915e7d659fb60ed692e850) with errors to get you started

japaric:
For a real life error (the SVD is wrong) see the [M36B.svd](https://github.com/posborne/cmsis-svd/blob/master/data/Toshiba/M36B.svd):

``` xml
    <peripheral>
      <name>UART0</name>
      <registers>
(..)
        <register>
          <name>MIS</name>
          <description>Masked Interrupt Status Register</description>
          <addressOffset>0x0040</addressOffset>
          <size>32</size>
          <access>read-only</access>
          <resetValue>0xfffffffffffffff1</resetValue>
```

`0xfffffffffffffff1` doesn't fit in 32 bits.

homunkulus:
:umbrella: The latest upstream changes (presumably 0989dbc) made this pull request unmergeable. Please resolve the merge conflicts.

Emilgardis:
I'll close this pr as it has gone stale. I will make a new pr eventually.

Emilgardis:
I've begun work on [here](https://github.com/emilgardis/svd/tree/feature/error-handling)

PR #29, opened 2017-04-30
Fixed underflow when width is 0

Emilgardis:
I'm not sure why the width is set as 0, but this prevents a panic

cc @whitequark


japaric:
`end > start` sounds a like a bug in the SVD file. I think it should be fixed there.

japaric:
Closing as I think this is a bug in the SVD file.

PR #34, opened 2017-06-20
add support for cpu element in SVD

ahixon:
I haven't added any of the optional elements from the "cpu" tag, since we don't really need them at the moment.

Use case for this pull request is another pull request I'll be submitting shortly on [svd2rust](/japaric/svd2rust). I use the CPU name tag to see if we're not a Cortex M device, and if so, skip core peripheral generation.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -70,6 +70,7 @@ impl ElementExt for Element {
 #[derive(Clone, Debug)]
 pub struct Device {
     pub name: String,
+    pub cpu: Option<CPU>,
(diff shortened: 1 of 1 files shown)


japaric:
Thanks @ahixon.

This looks correct as per the SVD specification. Let's see if all the SVD files in the test suite adhere to the spec ...

@homunkulus r+

homunkulus:
:pushpin: Commit 99bc675 has been approved by `japaric`

homunkulus:
:hourglass: Testing commit 99bc675 with merge 99bc675...

Issue #33, opened 2017-06-06
support <clusters>


japaric:
these add namespacing to registers and can be nested

estutzenberger:
I have been working on updates to svd for this. Please take a look at the following output and see if matches your expectation for how clusters should be processed.  If the output seems good, I'll push up the changes so that we can review.

A cluster example from nrf52.svd:
`
    <cluster>        
        <name>INFO</name>
        <description>Device info</description>
          <addressOffset>0x100</addressOffset>
          <register>
            <name>PART</name>
            <description>Part code</description>
            <addressOffset>0x000</addressOffset>
            <access>read-only</access>
            <resetValue>0x00052832</resetValue>
            <fields>
              <field>
                <name>PART</name>
                <description>Part code</description>
                <lsb>0</lsb>
                <msb>31</msb>
                <enumeratedValues>
                  <enumeratedValue>
                    <name>N52832</name>
                    <description>nRF52832</description>
                    <value>0x52832</value>
                  </enumeratedValue>
                  <enumeratedValue>
                    <name>Unspecified</name>
                    <description>Unspecified</description>
                    <value>0xFFFFFFFF</value>
                  </enumeratedValue>
                </enumeratedValues>
              </field>
            </fields>
          </register>
          <register>
            <name>VARIANT</name>
            <description>Part Variant, Hardware version and Production configuration</description>
            <addressOffset>0x004</addressOffset>
            <access>read-only</access>
            <resetValue>0x41414142</resetValue>
            <fields>
              <field>
                <name>VARIANT</name>
                <description>Part Variant, Hardware version and Production configuration, encoded as ASCII</description>
                <lsb>0</lsb>
                <msb>31</msb>
                <enumeratedValues>
                  <enumeratedValue>
                    <name>AAAA</name>
                    <description>AAAA</description>
                    <value>0x41414141</value>
                  </enumeratedValue>
                  <enumeratedValue>
                    <name>AAAB</name>
                    <description>AAAB</description>
                    <value>0x41414142</value>
                  </enumeratedValue>
                  <enumeratedValue>
                    <name>AABA</name>
                    <description>AABA</description>
                    <value>0x41414241</value>
                  </enumeratedValue>
                  <enumeratedValue>
                    <name>AABB</name>
                    <description>AABB</description>
                    <value>0x41414242</value>
                  </enumeratedValue>
                  <enumeratedValue>
                    <name>Unspecified</name>
                    <description>Unspecified</description>
[...]

PR #36, opened 2017-08-01
Implement processing of register clusters

estutzenberger:
See issue #33 for output.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -249,9 +249,15 @@ pub struct RegisterArrayInfo {
 }
 
 #[derive(Clone, Debug)]
+pub struct RegisterClusterInfo {
+    pub registers: Option<Vec<Register>>,
+}
+
+#[derive(Clone, Debug)]
 pub enum Register {
     Single(RegisterInfo),
     Array(RegisterInfo, RegisterArrayInfo),
+    Cluster(RegisterInfo, RegisterClusterInfo),
 }
 
 impl Deref for Register {
@@ -261,6 +267,7 @@ impl Deref for Register {
         match *self {
             Register::Single(ref info) => info,
             Register::Array(ref info, _) => info,
+            Register::Cluster(ref info, _) => info,
         }
     }
 }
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -303,11 +310,24 @@ impl RegisterArrayInfo {
...
+                        .filter(|c| c.name == "register")
+                        .filter_map(Register::parse)
+                        .collect())
+        }
+    }
+}
+
 impl Register {
     // TODO handle "clusters", return `Register` not an `Option`
     fn parse(tree: &Element) -> Option<Register> {
```
japaric: This should return `Register` since no branch returns `None` now.

src/lib.rs
```diff
@@ -303,11 +310,24 @@ impl RegisterArrayInfo {
     }
 }
 
+impl RegisterClusterInfo {
+    fn parse(tree: &Element) -> RegisterClusterInfo {
+        RegisterClusterInfo {
+            registers: Some(tree.children.iter()
+                        .filter(|c| c.name == "register")
```
japaric: I don't think this is correct. `<cluster>` trees can be nested. So both "register" and "cluster" should be allowed here.

src/lib.rs
```diff
@@ -249,9 +249,15 @@ pub struct RegisterArrayInfo {
...
 #[derive(Clone, Debug)]
+pub struct RegisterClusterInfo {
+    pub registers: Option<Vec<Register>>,
+}
+
+#[derive(Clone, Debug)]
 pub enum Register {
     Single(RegisterInfo),
     Array(RegisterInfo, RegisterArrayInfo),
+    Cluster(RegisterInfo, RegisterClusterInfo),
```
japaric: I think I would prefer a separate `Cluster` type instead of adding a variant here. Possibly with these fields: `struct Cluster { info: RegisterInfo, chidren: Either<Register, Cluster> }`.

`Register` should be used only for `<register>` trees.

src/lib.rs
estutzenberger: Ahh, yes, I see that in the spec now. Easy enough to patch up.

src/lib.rs
japaric: Yeah, I was thinking that the type of the `Peripheral.registers` would change to `Vec<Either<Cluster, Register>>`. That would reflect the actual SVD syntax where `<registers>` is an array of `<register>`s and `<cluster>`s.

> I am wondering then if it makes sense to create a new enum

We can use [either](https://crates.io/crates/either)

src/lib.rs
japaric: Then this could change to parse the *contents* of a `<register>` tree and there would be a *function* that parses either a `<register>` or a `<cluster>` that calls `Register::parse` or `Cluster::parse` depending on the tag.

PR #40, opened 2017-10-01
Added parsing of binary numbers with '0b' as a prefix

lucasbrendel:
I have an SVD file that is attached in japaric/svd2rust#146 that was failing parsing. I tracked it down to enumeratedValues having value tags with text in the format of '0b#' where '#' is a 0 or 1. I added an else clause to check for that and convert appropriately and keep the same 'don't care' syntax as the existing binary parsing.

There is still another issue with the SVD file, but it does not panic at least now at that stage.

### DIFF
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -10,6 +10,11 @@ pub fn u32(tree: &Element) -> Option<u32> {
         // #01101x1
         // along with don't care character x (replaced with 0)
         u32::from_str_radix(&str::replace(&text["#".len()..], "x", "0"), 2).ok()
+    } else if text.starts_with("0b"){
+        // Handle strings in the binary form of:
+        // 0b01101x1
+        // along with don't care character x (replaced with 0)
+        u32::from_str_radix(&str::replace(&text["0b".len()..], "x", "0"), 2).ok()
     } else {
         text.parse().ok()
     }


japaric:
Thanks for the PR, @lucasbrendel.

I was checking the [SVD specification](https://www.keil.com/pack/doc/CMSIS/SVD/html/schema_1_2_gr.html) and it seems that these "scaledNonNegativeInteger" fields only support the decimal and hexadecimal formats.

So your SVD file is actually not standard compliant. Which ... is not surprising because most vendor SVD files are not standard compliant either. In general, `svd-parser` is quite lenient in not enforcing some *restrictions* that the SVD spec mandates; this let us parse non standard SVD files.

I think this would be the first time we would deviate from the standard by adding *more parsing logic* but seems fine if we can support more SVD files.

@homunkulus r+

homunkulus:
:pushpin: Commit 5dc1f2d has been approved by `japaric`

homunkulus:
:hourglass: Testing commit 5dc1f2d with merge 5dc1f2d...

homunkulus:
:sunny: Test successful - [status-travis](https://travis-ci.org/japaric/svd/builds/282189332?utm_source=github_status&utm_medium=notification)
Approved by: japaric
Pushing 5dc1f2d731cfe609415c56b4b73dabb6507ec448 to master...

PR #43, opened 2017-10-17
Changing parse for '#' replace to match for both 'X' and 'x'

lucasbrendel:
This was found while testing other infineon XMC SVD files. From what i could tell, this should be allowed with the SVD schema, but correct me if this is allowing another edge case

### DIFF
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -9,12 +9,7 @@ pub fn u32(tree: &Element) -> Option<u32> {
         // Handle strings in the binary form of:
         // #01101x1
         // along with don't care character x (replaced with 0)
-        let s:String = text["#".len()..].chars()
-                    .map(|x| match x {
-                        'X' => '0',
-                        'x' => '0',
-                        _ => x}).collect();
-        u32::from_str_radix(&s, 2).ok()
+        u32::from_str_radix(&str::replace(&text.to_lowercase()["#".len()..], "x", "0"), 2).ok()
     } else if text.starts_with("0b"){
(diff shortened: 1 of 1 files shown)


src/parse.rs
```diff
@@ -9,7 +9,12 @@ pub fn u32(tree: &Element) -> Option<u32> {
         // Handle strings in the binary form of:
         // #01101x1
         // along with don't care character x (replaced with 0)
-        u32::from_str_radix(&str::replace(&text["#".len()..], "x", "0"), 2).ok()
```
Emilgardis: I'd prefer

```rust
&text.to_lowercase()["#".len()..]
```

lucasbrendel: That is a much better approach. I can adjust that.

japaric:
Thanks, @lucasbrendel. I'm not sure if this is actually allowed by the SVD spec but LGTM.

@homunkulus r+

homunkulus:
:pushpin: Commit fd1ce6a has been approved by `japaric`

homunkulus:
:hourglass: Testing commit fd1ce6a77d74b49988c5e8b075348945cf417211 with merge 53f25106e9260827039b518a17b8834979e8a682...

homunkulus:
:sunny: Test successful - [status-travis](https://travis-ci.org/japaric/svd/builds/290174644?utm_source=github_status&utm_medium=notification)
Approved by: japaric
Pushing 53f25106e9260827039b518a17b8834979e8a682 to master...

PR #44, opened 2018-02-21
Device: add all properties except vendorExtensions

jmesmon:
Seems like for `vendorExtensions` we'd need to expose an `Element`. Thoughts on that in a follow up PR?


jmesmon:
I'm going to close this as I'd prefer #37 and related not be delayed (this change is only useful to me if the ability to write out svds exists, anyhow).

Issue #46, opened 2018-03-06
The Refactor


ryankurte:
We have a few outstanding issues that to move forward with could use a refactor of the existing codebase.

## Issues:
- [svd#18](https://github.com/japaric/svd/issues/18) - Better error handling
   - [Emilgardis:feature/error-handling](https://github.com/japaric/svd/compare/master...Emilgardis:feature/error-handling?expand=1) - @Emilgardis WIP error handling branch
- [svd2rust#96](https://github.com/japaric/svd2rust/issues/96) - Common traits for identical IP blocks
   - [ryankurte:feature/encode](https://github.com/japaric/svd/pull/37) - My previous attempt at implementing encode/decode w/tests 
- [svd2rust#186](https://github.com/japaric/svd2rust/issues/186) - Proposal: User supplied scripts to add abstractions missing-from/impossible-in the input SVD.
    - should be covered by work on [svd2rust#96](https://github.com/japaric/svd2rust/issues/96) when we get this done
 
## Proposal
Things to do and PRs to accept. I'll update this list from any discussion in this issue to reflect our plan and progress.

### Project Level
1. [x] Introduce `Error` types for svd decoding (@Emilgardis?)
2. [x] Introduce `Encode` and `Decode` traits for svd objects (@ryankurte?)
3. [x] Refactor all svd objects (anyone and everyone)

Once we've defined  Error types and Encode/Decode traits the refactoring part should be reasonably simple, ~and we can accept a PR for each to avoid nightmare refactoring~.

### For each SVD object (ie. object implementing an SVD element)
1. [x] Move from `src/lib.rs` to `srv/svd/object.rs` and add `pub mod object;` to `src/svd/mod.rs`
2. [x] Update decode method to match `Decode` trait (and add error handling)
3. [ ] Create encode method to match `Encode` trait ([svd#37](https://github.com/japaric/svd/pull/37) has many pre-written)
4. [ ] Add tests for encoding and decoding each variant for an object.

~I propose that we put a freeze on changes to master while undertaking this and make these changes against a `refactor` branch until we reach stability again, then merge back to master in what is bound to be a spectacularly breaking change.~

Emilgardis:
Looks good, I have some thoughts however.
First, I don't think a new branch is needed for refactoring. 

I also believe making the structure of the project `src/svd/object.rs` is a better idea than having loose files in `src/`

Also, just to be clear, when you say SVD object, you mean the structs in `lib.rs` relating to elements in the SVD spec right?

----------

[...]

Issue #18, opened 2017-01-08
Better error handling


Emilgardis:
See related issue japaric/svd2rust#51
`svd-parser` uses `Option<String>` for `ElementExt::get_child_text`, and an expect is done, avoiding any way of knowing what had gone wrong. I propose for `svd-parser` to use `error-chain` to fix this problem, but also to use a logging crate to make debugging svd errors easier for the user without having to use utilities like `SVDConv.exe` to check for errors in the svd file and to see if the error is by missing implementation in `svd2rust`.

Emilgardis:
I have a WIP [branch]( https://github.com/Emilgardis/svd/commits/add_error_checking) that does this. It is not exactly what I would like it to be currently, but it gives me a much better picture of what happened on an error.

Emilgardis:
I'm going to wait for japaric/svd2rust#53 and then rebase my branch. I have also implemented #21 on that branch, it should be on a separate pr.

nickray:
@Emilgardis: running into this exact issue when trying to `svd2rust` the ARM Musca-B1 SVD. Any chance your branch is still rebaseable?

robamu:
I have dug a little bit deeper and I can confirm that the error comes from the `syn`  crate when calling `generate::device::render`. The precise error happens at `convert_svd_register` / `name_to_wrapped_ty`

PR #47, opened 2018-03-06
Implement error handling

Emilgardis:
This PR adds error handling via `failure`. This implementation does however need some work on it as the generated errors doesn't seem matchable in it's current state.

### TODO
1. Change `fn(...) -> Result<_,Error>` into `fn(...) -> Result<_,FnSpecificError>`
    This should fix item 2, but I'm not sure.
2. Make matching more ergonomic and always possible
3. Add more tests

Part of #46 and #18


ryankurte:
I'm not 💯 on implementing errors yet, but would the solution to match be to expose an `SVDError` enum to use that collects the implemented error types?

Emilgardis:
This may all be due to bad usage of `failure`, but an example of not being able to match on errors is [this](https://github.com/Emilgardis/svd/blob/cfe1de055e866a24248dc7764132801e2e3867e5/tests/bad_svd.rs#L43) ignored test.

I am currently looking into implementing `failure::Error` manually instead of using the `#[derive(Fail)]`.

Emilgardis:
Work is being reimplemented in the [refactor](https://github.com/japaric/svd/tree/refactor) branch

PR #48, opened 2018-03-07
Added Encode/Decode traits (and new_element helper)

ryankurte:
Traits for encoding and decoding of SVD objects <-> XML Elements towards #46.
@Emilgardis I think this should be compatible with your approach to error handling?

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -1,0 +1,33 @@
+// Helper traits for rust-svd
+
+use std::collections::HashMap;
+use xmltree::Element;
+
+/// Parse trait allows SVD objects to be parsed from XML/SVD elements.
+pub trait Parse {
+    /// Object returned by parse method
+    type Object;
+    /// Parsing error
+    type Error;
+    /// Parse an XML/SVD element into an SVD Object.
+    fn parse(&Element) -> Result<Self::Object, Self::Error>;
+}
+
+/// Encode trait allows SVD objects to be encoded into XML/SVD elements.
+pub trait Encode {
+    /// Encoding error
+    type Error;
+    /// Encode an SVD object into an XML element
+    fn encode(&self) -> Result<Element, Self::Error>;
+}
+
+/// new_element helper to create new xml elements
+pub fn new_element(name: &str, text: Option<String>) -> Element {
+    Element {
+        name: String::from(name),
+        attributes: HashMap::new(),
+        children: Vec::new(),
+        text: text,
+    } 
+}
+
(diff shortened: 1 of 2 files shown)


Emilgardis:
Yes it should work fine, however I don't see the point in this pr. I think adding everything together makes more sense.

I can redo my work on error-handling, that's fine.

src/types.rs
```diff
@@ -0,0 +1,33 @@
...
+use std::collections::HashMap;
+use xmltree::Element;
+
+/// Parse trait allows SVD objects to be parsed from XML/SVD elements.
+pub trait Parse {
+    /// Object returned by parse method
+    type Object;
+    /// Parsing error
+    type Error;
+    /// Parse an XML/SVD element into an SVD Object.
```
Emilgardis: Sentence doesn't feel right to me. Maybe 

> [...] into it's corresponding `Object`

src/types.rs
```diff
@@ -0,0 +1,33 @@
...
+    fn parse(&Element) -> Result<Self::Object, Self::Error>;
+}
+
+/// Encode trait allows SVD objects to be encoded into XML/SVD elements.
+pub trait Encode {
+    /// Encoding error
+    type Error;
+    /// Encode an SVD object into an XML element
```
Emilgardis: Same here as above. Maybe just skip "an SVD object".

src/types.rs
```diff
@@ -3,21 +3,21 @@
...
 
-/// Parse trait allows SVD objects to be parsed from XML/SVD elements.
+/// Parse trait allows SVD objects to be parsed from XML elements.
 pub trait Parse {
     /// Object returned by parse method
     type Object;
     /// Parsing error
     type Error;
-    /// Parse an XML/SVD element into an SVD Object.
+    /// Parse an XML/SVD element into it's corresponding Object.
```
Emilgardis: Object refers to the trait type, put into code block. i.e `` `Object` ``

Emilgardis:
I still feel uneasy about merging this without anything related to it. Maybe it would actually be easier to open a refactor branch and just directly push to it and then merge it, as you first suggested.

ryankurte:
If you'd like I could update a couple of objects and open dependent PRs as
well so we can test / prove the interface? Then it's easy to merge em also.

Emilgardis:
I say yes to implementing this on some objects, but no to opening dependent PRs. If we are going to make multiple small PRs depending on this I think skipping the PR system is the best choice.

ryankurte:
Closed in favour of [upstream/refactor](https://github.com/japaric/svd/tree/refactor)

Issue #21, opened 2017-02-05
Implement derivedFrom on cluster/register/field/enumeratedvalues element


Emilgardis:
**Repurposing this issue for all `derivedFrom` implementations.**
----------
The SVD spec specifies a `derivedFrom` attribute on clusters, registers, fields and enumeratedvalues.
To be able to use this attribute we need to supply this information to crates using this library.

PR #49, opened 2018-03-11
Add a function to produce merged peripheral when derivedFrom is used.

nsabovic:
Added a function to make a new peripheral by overlaying the peripheral over its derivedFrom peripheral. I read the spec that it is a shallow copy, but I could be wrong. This may be part of #21.

With this change, and another one that I have for svd2rust, svd2rust can generate peripherals with derivedFrom. I could have put this function over there, but I think the way overlaying is done is a part of the SVD spec, and not part of the particular svd2rust implementation, so it makes more sense for me to put it here.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -178,6 +178,23 @@ pub struct Peripheral {
 }
 
 impl Peripheral {
+    pub fn derive_from(&self, other: &Peripheral) -> Peripheral {
+        let mut derived = self.clone();
+        if derived.group_name.is_none() && other.group_name.is_some() {
+            derived.group_name = other.group_name.clone();
+        }
+        if derived.description.is_none() && other.description.is_some() {
+            derived.description = other.description.clone();
+        }
+        if derived.registers.is_none() && other.registers.is_some() {
+            derived.registers = other.registers.clone();
+        }
+        if derived.interrupt.is_empty() {
+            derived.interrupt = other.interrupt.clone();
+        }
+        derived
+    }
+
     fn parse(tree: &Element) -> Peripheral {
         assert_eq!(tree.name, "peripheral");
 


Emilgardis:
I'm not sure about having this in this crate. What reasons are there for it being in this crate instead of having it as a function in `svd2rust`?

nsabovic:
Could just as well be there, if this crate is just to be tasked with parsing. However, if someone writes `svd2c++`, they will also need this function. It is only called once in `svd2rust`, it would be easy to stick it there, if that is your design preference.

src/lib.rs
```diff
@@ -178,6 +178,23 @@ pub struct Peripheral {
 }
 
 impl Peripheral {
+    pub fn derive_from(&self, other: &Peripheral) -> Peripheral {
+        let mut derived = self.clone();
+        if derived.group_name.is_none() && other.group_name.is_some() {
```
wez: minor nit: the conditionals here don't need to check `other.XXX`.   If they're `is_none` then the `derived` value will get assigned to `None` anyway.  It just feels a little less repetitive:

```
  if derived.group_name.is_none() {
       derived.group_name = other.group_name.clone();
  }
```

Emilgardis: I second this. I'll merge this when this has been fixed.

Emilgardis:
@nsabovic I took the liberty and implemented this in #52.

nsabovic:
@Emilgardis —Thanks for the patience, and for fixing up my awkward style!

The fixed version has lines like:

    derived.group_name = derived.group_name.or(other.group_name.clone())

Shouldn't it be:

    derived.group_name = derived.group_name.or_else(|| other.group_name.clone())

To avoid calling `clone()` unconditionally, and then discarding it if derived already has the value?

Emilgardis:
@nsabovic Yes I should have done that. Right now master is in a frozen state as we are working on #46, I fixed this in bdfa727 however. Thank you for mentioning it! :+1:

PR #50, opened 2018-03-13
Parse <register><alternateRegister> elements

wez:
In looking at https://github.com/japaric/svd2rust/issues/191 and https://github.com/japaric/svd2rust/issues/16 I thought that this might help, so here's the trivial patch.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 *.rs.bk
+[._]*.sw[a-p]
 Cargo.lock
 target
 tests/
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -239,7 +239,9 @@ pub struct ClusterInfo {
 #[derive(Clone, Debug)]
 pub struct RegisterInfo {
     pub name: String,
+    pub alternate_group: Option<String>,
     pub alternate_register: Option<String>,
+    pub derived_from: Option<String>,
     pub description: String,
     pub address_offset: u32,
     pub size: Option<u32>,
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -1,4 +1,5 @@
 *.rs.bk
+.*sw*
```
Emilgardis: `[._]*.sw[a-p]` is a better solution for matching vim swap files.

src/lib.rs
```diff
@@ -351,6 +354,13 @@ impl RegisterInfo {
     fn parse(tree: &Element) -> RegisterInfo {
         RegisterInfo {
             name: try!(tree.get_child_text("name")),
+            alternate_group: tree.get_child_text("alternateGroup"),
+            alternate_register: tree.get_child_text("alternateRegister"),
+            derived_from: tree.attributes.get("derivedFrom").map(
+                |s| {
+                    s.to_owned()
+                },
+            ),
```
Emilgardis: This indentation is uneeded

```rust
tree.attributes.get("derivedFrom").map(|s| s.to_owned()),
```

.gitignore
wez: will amend

src/lib.rs
wez: I copied this from the other instance of `derivedFrom`.  Note that the `.rustfmt.toml` config wants this to be:

```
            derived_from: tree.attributes
                .get("derivedFrom")
                .map(|s| s.to_owned()),
```

There are other portions that are not rustfmt'd.   Would you like me to insert a commit that applies rustfmt underneath my changes and then make them match?

wez:
https://github.com/japaric/svd/pull/51 for rustfmt, I'll amend this one independently from that for now

wez:
(force pushed with your suggested changes; you'll probably need to hard refresh the PR page)

Emilgardis:
Thank you!

bors r+

PR #53, opened 2018-03-21
Refactor

ryankurte:
Work In Progress PR for SVD refactor (#46) so that we get CI build/test going.


Emilgardis:
I've made a page for tracking what elements we currently support https://github.com/japaric/svd/wiki/Compatibility-with-SVD-spec

Emilgardis:
There are some usages of `SVDError::Other` which should be made into proper enums, but I feel that is not very important.

ryankurte:
I wasn't sure whether to do it everywhere or just at the top level, this seems better.
Thanks / lgtm!

ryankurte:
Oh oops, should I have used bors to do this..? :-/

japaric:
@ryankurte yes, because we don't run CI on PRs. But it's OK, if you haven't break the build.

ryankurte:
It looks like we don't currently build master either?
screen shot 2018-09-03 at 3 32 00 pm

As an aside, is there a reason for not running CI automatically / all the time (even alongside bors)? IMO it's a useful indicator, and aligns with expectations of how github/travis works for people new to rust (myself definitely included).

japaric:
> It looks like we don't currently build master either?

Yep, because it's redundant with what bors tests before landing the PR.

>  is there a reason for not running CI automatically / all the time

I have way too many repos and CI capacity (about 5 parallel builds) is shared across all of them. In my case, having a build trigger everything time PR is open or updated severely bottlenecked landing PRs using bors so I disabled PR builds across the board long time ago (before that I had to manually kill some still under review PR builds to land approved PRs).

I have PR builds enabled on the rust-embedded repos (\*) because the org has its own build capacity (also 5 parallel builds) and there aren't too many repos there.

(\*) Except for Cross because the build time is longer than 1 hour and has many jobs. If it was enabled each build would block the bors queue for at least 1 hour.

ryankurte:
> I have way too many repos and CI capacity

Ahh, that's a very relatable problem ^_^

> I have PR builds enabled on the rust-embedded repos

I appreciate the change on rust-embedded! Is the plan to move this over at some point or keep it here?

PR #54, opened 2018-05-02
Ignore UTF-8 BOM from the xml input

ryochack:
This PR is to support the SVD input that contains UTF-8 BOM. (Ignore BOM)
Please refer to [the svd2rust's PR](https://github.com/japaric/svd2rust/pull/200).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -130,7 +130,7 @@ impl ElementExt for Element {
 
 /// Return the &str trimmed UTF-8 BOM if the input &str contains the BOM.
 fn trim_utf8_bom(s: &str) -> &str {
-    if s.len() > 2 && s.as_bytes().starts_with(&[0xefu8, 0xbbu8, 0xbfu8]) {
+    if s.len() > 2 && s.as_bytes().starts_with(b"\xef\xbb\xbf") {
         &s[3..]
     } else {
         s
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -128,9 +128,26 @@ impl ElementExt for Element {
...
-
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::str;
+
+    #[test]
+    fn test_trim_utf8_bom_from_str() {
+        // UTF-8 BOM + "xyz"
+        let bom_str = str::from_utf8(&[0xefu8, 0xbbu8, 0xbfu8, 0x78u8, 0x79u8, 0x7au8]).unwrap();
```
Emilgardis: Same as earlier comment, could you use `b"\xef\xbb\xbfxyz"`

src/lib.rs
```diff
@@ -128,9 +128,26 @@ impl ElementExt for Element {
     }
 }
 
+/// Return the &str trimmed UTF-8 BOM if the input &str contains the BOM.
+fn trim_utf8_bom(s: &str) -> &str {
+    if s.len() > 2 && s.as_bytes().starts_with(&[0xefu8, 0xbbu8, 0xbfu8]) {
```
Emilgardis: I would prefer `b"\xef\xbb\xbf"` here instead of the array

ryochack:
Thanks for your advice!
I fixed 31e09842.

Emilgardis:
Thank you!
merging without CI as this is still wip

PR #55, opened 2018-06-05
Made the u32 parse use the Parse trait

VictorKoenders:
I was going to change the `fn u32(&Element)` to `fn u32(&str)` like the TODO says [here](https://github.com/japaric/svd/blob/refactor/src/parse.rs#L15) but then I noticed that the rest of the codebase uses [`parse::optional`](https://github.com/japaric/svd/blob/refactor/src/parse.rs#L63).

Because `parse::optional` expects `CB: 'static + Fn(&Element) -> Result<T, SVDError>`, I decided to just implement `impl Parse for u32`

This also allows us to change the `parse::optional` definition to:
```rust
pub fn optional<'a, T>(n: &str, e: &'a Element) -> Result<Option<T::Object>, SVDError>
    where T : Parse<Error = SVDError>
{
     let child = match e.get_child(n) {
        Some(c) => c,
        None => return Ok(None),
    };

    match CB::parse(child) {
        Ok(r) => Ok(Some(r)),
        Err(e) => Err(e),
    }
}
```
which would change the function call from:
```rust
parse::optional("size", tree, u32::parse)?;
```
to one of:
```rust
parse::optional::<u32>("size", tree)?;
parse::optional("size", tree)?;
```

( I'm willing to change the parse::optional implementation if that's something we want)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -107,7 +107,7 @@ impl ElementExt for Element {
     /// Get a u32 value from a named child element
     fn get_child_u32(&self, n: &str) -> Result<u32, SVDError> {
         let s = self.get_child_elem(n)?;
-        parse::u32(&s).context(SVDErrorKind::ParseError(self.clone())).map_err(|e| e.into())
+        u32::parse(&s).context(SVDErrorKind::ParseError(self.clone())).map_err(|e| e.into())
     }
 
     /// Get a bool value from a named child element
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -1,35 +1,10 @@
 use xmltree::Element;
-use failure::ResultExt;
-use ElementExt;
 
 use error::*;
 
 macro_rules! try {
     ($e:expr) => {
         $e.expect(concat!(file!(), ":", line!(), " ", stringify!($e)))
-    }
-}
-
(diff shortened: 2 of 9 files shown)


VictorKoenders:
As for the idea to simplify parse::optional to use generics, turns out Rust can't quite figure out what implementation of `Parse` to use, so you'd have to end up specifying it every time.

I guess this is a matter of preference. https://gist.github.com/VictorKoenders/3f262de44e51fca70112cf85decb2af1

ryankurte:
Thanks for the PR! Looks like an improvement / good to me, though I guess we should convert `bool` at the same time. Any thoughts @Emilgardis?

Emilgardis:
I think it looks good. :+1:

ryankurte:
I'm happy to accept this here, thanks for the help @VictorKoenders!
I would also definitely support further PRs to:
- move to your proposed `parse::optional::<u32>("size", tree)?;` approach
- make the same changes to [bool](https://github.com/japaric/svd/blob/979cf99d6dc66c38e91b4a0031dfcbbd5c5ba3cd/src/parse.rs#L36)

If you're interested in working on them ^_^

PR #56, opened 2018-06-09
Applied requested changes to parse::optional

VictorKoenders:
Applied the requested changes in #55.

As it turns out, `parse::optional` does not get used in combination with `bool`.

I have made a `BoolParse` struct that implements parse, but because `pub fn bool(tree: &Element)` returned `Option<bool>`, this makes `parse::optional` return `Option<Option<bool>>`.

One way to fix this is to put more logic in the `Parse` trait that returns an optional by default, like so:

```rust
/// Parse trait allows SVD objects to be parsed from XML elements.
pub trait Parse {
    /// Object returned by parse method
    type Object;
    /// Parsing error
    type Error;
    /// Parse an XML/SVD element into it's corresponding `Object`.
    fn parse(&Element) -> Result<Self::Object, Self::Error>;

    /// Parse an XML/SVD element into it's corresponding `Option<Object>`.
    fn parse_optional(&Element) -> Result<Option<Self::Object>, Self::Error> {
        match Self::parse(child) {
            Ok(r) => Ok(Some(r)),
            Err(e) => Err(e),
        }
    }
}
```

In this situation, the `BoolParse` struct can overwrite the `parse_optional` fn to make sure the `Option` does not get doubled.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -41,6 +41,8 @@ pub enum SVDErrorKind {
     InvalidRegisterCluster(Element, String),
     #[fail(display = "Invalid modifiedWriteValues variant, found {}", _1)]
     InvalidModifiedWriteValues(Element, String),
+    #[fail(display = "The content of the element could not be parsed to a boolean value {}: {}", _1, _2)]
+    InvalidBooleanValue(Element, String, ::std::str::ParseBoolError),
     #[fail(display = "encoding method not implemented for svd object {}", _0)]
     EncodeNotImplemented(String),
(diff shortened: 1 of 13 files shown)


ryankurte:
Thanks for the PR! I think it would make more sense to update parse for `BoolParse` to return a `Result<bool, SVDError>` and use the `parse_optional` method if/where it's required so the interfaces are consistent / explicit, unless I'm missing something?

ryankurte:
¯\_(ツ)_/¯ well that doesn't work, will have to find out why.

Thanks for the PR!

PR #61, opened 2018-11-15
encode/field: only generate a description element when present

japaric:
without this parsing the output of encode fails when there is no description
field

the first commit can be ignored as it just runs cargo-fmt


japaric:
rebased and dropped the `cargo fmt` commit

PR #64, opened 2018-11-27
Make register description optional as per svd spec

### LINKED ISSUE #59: Should register description should be optional?
I recently ran into an SVD that has single tag empty descriptions for a few register clusters (ATSAML11). It seems like that is [against the SVD spec](http://www.keil.com/pack/doc/CMSIS/SVD/html/svd_xml_conventions_gr.html#_), so I manually removed the tag entirely. Unfortunately that still doesn't parse correctly, as the description field is currently required.

If I'm reading http://www.keil.com/pack/doc/CMSIS/SVD/html/elem_registers.html correctly though it sounds like the description field should be optional.

Emilgardis:
Fixes #59

### DIFF
--- a/src/svd/registerinfo.rs
+++ b/src/svd/registerinfo.rs
@@ -24,7 +24,7 @@ pub struct RegisterInfo {
     pub alternate_group: Option<String>,
     pub alternate_register: Option<String>,
     pub derived_from: Option<String>,
-    pub description: String,
+    pub description: Option<String>,
     pub address_offset: u32,
     pub size: Option<u32>,
     pub access: Option<Access>,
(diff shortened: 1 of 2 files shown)


src/svd/registerinfo.rs
```diff
@@ -105,15 +105,22 @@ impl Encode for RegisterInfo {
...
                     "addressOffset",
                     Some(format!("0x{:x}", self.address_offset)),
                 ),
             ],
             text: None,
         };
-
+        match self.description {
+            Some(ref v) => {
+                elem.children.push(new_element(
```
Emilgardis: This will fail ci. This is because the encode tests are dependent on the order of the children vec. To fix this we can either make the test order agnostic with a sort on all element children or use an insert.
I'm not sure which one I prefer, as making the test order agnostic seems like a more permanent solution should this issue arise again

```suggestion
                elem.children.insert(1, new_element(
```

ryankurte: Order agnostic sounds good to mee!

Emilgardis:
Implemented order agnostic tests for encode

ryankurte:
I think this nightly error was fixed in #65

bors retry

ryankurte:
fixed conflict and tested locally

bors retry

ryankurte:
hmmm

bors r+

PR #65, opened 2018-12-08
derivedFrom is an attribute, not a child of <register>

wez:
This got accidentally refactored to be parsed as a child,
but this is actually an attribute (consistent with peripheral.rs).

Fixup and test via `cargo test --features unproven`

See also: https://github.com/rust-embedded/svd2rust/pull/256

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -26,7 +26,6 @@
 
 extern crate either;
 extern crate xmltree;
-#[macro_use]
 extern crate failure;
 
 #[cfg(feature = "unproven")]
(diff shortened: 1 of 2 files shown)


Emilgardis:
Thanks for the pr!

bors r+

ryankurte:
@wez we have a deadlock with this and removing the unused `#[macro_use]` that's causing build failure on new stable/nightly. Could you add that please, then tests should pass here.

(I tried myself, but don't have permissions to edit PRs on this repo)

wez:
This PR doesn't use change any use of anything from the failure crate in the travis build failure.  The same failure occurs in master.  I can put that change in this PR, but it is completely unrelated.

Emilgardis:
Thanks for patching it up anyway 👍 

bors r+

PR #67, opened 2019-03-03
Support for optional bitWidth in bit ranges

Disasm:


Disasm:
In SVD schema (v1.1, v1.2, v1.3) `bitWidth` field is marked as optional.

src/svd/bitrange.rs
```diff
@@ -91,21 +91,23 @@ impl Parse for BitRange {
...
-                    width: u32::parse(width).context(SVDErrorKind::InvalidBitRange(tree.clone(), InvalidBitRange::ParseError))?, 
-                    range_type: BitRangeType::OffsetWidth
-                })
-            )
+        } else if let Some(offset) = tree.get_child("bitOffset")
+        {
+            let width = if let Some(width) = tree.get_child("bitWidth") {
+                u32::parse(width).context(SVDErrorKind::InvalidBitRange(tree.clone(), InvalidBitRange::ParseError))?
+            } else {
+                1
```
japaric: Please add a comment here quoting or linking the part of the spec that says that the bit width is "1" when omitted.

Disasm: I tried hard to find this in docs, but failed. This "default=1" assumption was made based on common sense and some 3rd-party parser implementation. I have just tried an official ["SVDConv.exe"](https://github.com/ARM-software/CMSIS/blob/master/CMSIS/Utilities/SVDConv.exe) checker/generator and it treats absent `bitWidth` as an error. Moreover, for `<bitOffset>10</bitOffset>` field it produces the following output:
```
test.svd(73) : error [Level: <fields>]
MSB not set for Field: 'RST', default to '0'
Tag:       [TAG_END] field

test.svd(73) : error [Level: <fields>]
LSB > MSB for 'RST'. Set to [10..0]
Tag:       [TAG_END] field
```
Now it's clear for me that we have absolutely no agreement on the behavior of the `bitOffset` bitfield without `bitWidth` defined.

PR #69, opened 2019-05-22
Fix enumeratedValues parsing

Disasm:
Checks that `enumeratedValues` elements do not contain unexpected tags.
Also checks that `enumeratedValues` contains at least one `enumeratedValue`.

### DIFF
--- a/src/svd/enumeratedvalues.rs
+++ b/src/svd/enumeratedvalues.rs
@@ -44,15 +44,29 @@ impl Parse for EnumeratedValues {
             values: {
                 let values: Result<Vec<_>, _> = tree.children
                     .iter()
-                    .filter(|t| t.name == "enumeratedValue")
+                    .filter(|t| ["name", "headerEnumName", "usage"].iter().all(|s| &t.name != s))
                     .enumerate()
                     .map(|(e, t)| {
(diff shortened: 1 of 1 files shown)


Disasm:
Fixed the warnings. I'm not sure if `// TODO: is this correct? probably not, need tests` comment is obsolete, maybe @ryankurte could say something.

Disasm:
bors try

Emilgardis:
bors try

Disasm:
Fixed.

Emilgardis:
bors try

burrbull:
It would be good also to check if <enumeratedValues> contains items with identical names or values.

Disasm:
@burrbull, agree, but this semantic information check matches `svd2rust` better. Added an issue for this.

burrbull:
@Disasm can you resolve conflicts?

Disasm:
Done!
I excluded fix of these lines: https://github.com/rust-embedded/svd/blob/d7464ff910afa321717b1a97bf46afc13a6148a0/src/svd/register.rs#L65-L68
Maybe it's better to add it with a separate change.

burrbull:
bors try

PR #74, opened 2019-08-10
Peripheral Cluster Defaults

burrbull:
Add register ~~`Defaults`~~ `RegisterProperties` for Peripheral and ClusterInfo.

https://www.keil.com/pack/doc/CMSIS/SVD/html/elem_peripherals.html

PR #75, opened 2019-08-14
Show unknown access variant

nickray:
The ARM Musca-B1 SVD has access "read", cf. https://github.com/rust-embedded/svd2rust/issues/376

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -26,8 +26,8 @@ pub enum SVDErrorKind {
     ParseError(Element),
     #[fail(display = "NameMismatch")]
     NameMismatch(Element),
-    #[fail(display = "unknown access variant found")]
-    UnknownAccessType(Element),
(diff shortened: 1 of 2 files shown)


nickray:
Can you trigger bors again @ryankurte?

ryankurte:
:+1:

bors retry

ryankurte:
this should be fixed in #74

nickray:
@ryankurte I rebased on master, hope third time's the charm.

nickray:
By the way, how am I supposed to test if patches of mine work? The latest build failure was not at all visible to me pre-push via `cargo build --all-features`.

ryankurte:
yeeah, it's a bit tricky to test against a moving target :-/ you _can_ run the [test script](https://github.com/japaric/svd/blob/master/ci/script.sh) (which invokes cross rather than cargo) directly which would hopefully correlate slightly better.

thanks for your effort!

bors retry

PR #76, opened 2019-08-27
Add a 'serde_svd' feature which adds serde derive macros.

WRansohoff:
I found myself wanting to use `svd2rust` to parse SVD files into a JSON format, but it looks like this underlying `svd` crate doesn't support the `serde` Serialize / Deserialize macros which are used by a variety of Rust parsers and formatters.

Adding conditional `#[derive(Deserialize, Serialize)]` attributes to this crate seemed easier than creating a new crate with duplicate structs, and I put it behind an optional feature because most use cases won't require `serde` support.

Anyways, no worries if this addition is not within the scope of this crate, but I thought I'd share since it seems to work. Example project:

https://github.com/WRansohoff/svd2json

Thank you for all the work that you've put into the embedded Rust ecosystem! And sorry if the syntax needs cleaning up, or if there's a better way to do this - I don't have much experience with packaging Rust crates for release.

PR #80, opened 2019-09-05
Serde

burrbull:
Rebased and shortened #76

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,8 +11,14 @@ version = "0.7.0"
 
 [features]
 unproven = []
+serde_svd = ["serde"]
 
 [dependencies]
 either = "1.5"
 xmltree = "0.8"
 failure = "0.1"
+
+[dependencies.serde]
+version = "1.0"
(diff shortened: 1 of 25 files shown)


burrbull:
bors try

Cargo.toml
```diff
@@ -11,8 +11,14 @@ version = "0.7.0"
 
 [features]
 unproven = []
+serde_svd = ["serde"]
```
reitermarkus: Just call this `serde`.

burrbull: I've deleted this feature as `serde` is option.

burrbull:
Rebased once more.

burrbull:
bors try

therealprof:
Looks good to me per se but what is the benefit of having serde de-/serialisation here?

burrbull:
Added https://github.com/WRansohoff/svd2json as example.

therealprof:
Oooh, I see. That could be handy for a number of things...

PR #81, opened 2019-09-05
DerivedFrom trait

burrbull:


burrbull:
bors try

Disasm:
Are you sure this should be in svd *parser*?

burrbull:
Not, I'm not sure. But it should not be in `svd2rust` too.

Disasm:
Why do you think so? Are there any other "users" of this high-level functionality?

burrbull:
Rebased on #83

burrbull:
I located all `derive_from()` in separate file under the feature.

It will be easier to move this in `svd2rust` or other crate if we decide.

burrbull:
@therealprof  can we make new release with this PR?

therealprof:
@burrbull Can you remind me what these changes do and how to use them? Especially since they're behind a feature gate that might some some interesting implications.

burrbull:
It is needed to unify copying of 'derived' items (peripherals, registers, fields, etc.).
Now this functionality partially in in `svd-parser` and partially in `svd2rust` (https://github.com/rust-embedded/svd2rust/blob/11e11f0498c0aca1042eb927e2aa74c246a8b4a0/src/generate/peripheral.rs#L100).

Also https://github.com/rust-embedded/svd2rust/pull/372 requires it.
I've moved this under feature because this does not apply directly to parsing.

therealprof:
Okay, but before releasing we should properly modernize this crate, meaning:

1. Bump the dependencies
2. Turn it into edition 2018
3. Update the including the repository URL
4. Touch up the CHANGELOG.md to match the previous versions, bump the version and release it

PR #82, opened 2019-09-05
if let instead of match, rm ref

burrbull:


burrbull:
bors try

src/svd/enumeratedvalue.rs
```diff
@@ -75,31 +75,21 @@ impl Encode for EnumeratedValue {
             text: None,
         };
         // FIXME: Use if let some pattern here
```
Disasm: Remove the comment?

src/svd/register.rs
```diff
@@ -25,9 +25,9 @@ impl Deref for Register {
     type Target = RegisterInfo;
 
     fn deref(&self) -> &RegisterInfo {
-        match *self {
-            Register::Single(ref info) => info,
-            Register::Array(ref info, _) => info,
+        match &*self {
```
Disasm: Is `&*` required here?

src/svd/enumeratedvalue.rs
burrbull: Done.

src/svd/register.rs
burrbull: Removed.

Disasm:
Please, do not force-push changes after review, it will make it easier to review incrementally (instead of starting from scratch).

PR #88, opened 2019-09-08
remove `_extensible`

burrbull:
I could not find any sense in `_extensible` field.

It adds a lot of problems with constructing structs and I don't see any way where it can be usefull.

cc @rust-embedded/tools 

cc @japaric can you explain why you add this? Is this still needed?


Emilgardis:
The extensible field is used for future proofing. Its a "hack" until rust-lang/rust#44109 is implemented.

burrbull:
Closed in favour of #93.

PR #90, opened 2019-09-09
Field arrays

burrbull:
Support field arrays same way as `Register` and `Cluster`.

### Alternative:
Add `Option<DimElement>` in `Field`:
https://github.com/rust-embedded/svd/commit/c5ed1dc89b6c3b1cade1de3751f34d0b72d13101


burrbull:
bors try

burrbull:
It is surprise for me, but due to `Deref` this change doesn't break `svd2rust` compile.

PR #98, opened 2019-11-14
use `anyhow` as Error type instead of `failure`

burrbull:
https://crates.io/crates/anyhow + https://crates.io/crates/thiserror


burrbull:
bors r=Emilgardis

PR #102, opened 2019-12-20
Remove _extensible fields from structs.

jeandudey:
It limits the creation of these structs from the user point of view,
adding a field to an struct is always a breaking change so it doesn't
matter really if the user can construct it or not manually. This should
enable the usage of the encode API.

This doesn't break svd2rust imho.


HarkonenBade:
An alternative option would be to derive `Default` for these structures. As then someone trying to instantiate a struct can use the following form:
```
let s = Peripheral {
    name: String::from("foo"),
    base_address: 0xdeadbeef,
    ..Default::default()
};
```

burrbull:
In #88 we decided to use builders to solve this problem.
See #101 .

PR #103, opened 2020-01-02
ignore empty optional tag when get text

burrbull:
r? @therealprof 

This should fix faults on empty optional text tags like `<description/>`.


burrbull:
bors try

PR #115, opened 2020-04-24
Write Builders manually 2

burrbull:
#101  without changes in Error hierarchy.

Should slightly be easier to review.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -92,6 +92,10 @@ pub(crate) fn new_element(name: &str, text: Option<String>) -> Element {
     }
 }
 
+pub trait Build {
+    type Builder;
+}
+
 /// Generic test helper function
 /// Takes an array of (item, xml) pairs where the item implements
 /// Parse and Encode and tests object encoding and decoding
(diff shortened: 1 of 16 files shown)


burrbull:
bors try

src/lib.rs
```diff
@@ -92,6 +92,10 @@ pub(crate) fn new_element(name: &str, text: Option<String>) -> Element {
     }
 }
 
+pub trait Build {
+    type Builder;
+}
+
```
Emilgardis, line 98: Why is this not

```suggestion
pub trait Builder {
    type Output;
    type Error: std::error::Error + std::fmt::Display + 'static; // Maybe Send + Sync even?
    pub fn build(self) -> Result<Self::Output, Self::Error>;
}
```
and then move every impl of fn build to the trait. I don't see the reason for having this trait unless it is used like this.

burrbull: I removed this trait as it is not important. We can discuss this in future.

burrbull:
bors try

burrbull:
Done.

PR #120, opened 2020-05-10
Enforce register value constraints

gkelly:
Enforce register size, reset, and mask value agreement. If the provided
reset value or mask extend past the register size, or the provided reset
value would conflict with the mask, then fail parsing.

---

I saw that there are some tests from CMSIS that are disabled because of constraint violations due to a check like this. There are also a couple of ATSAMD51 tests (at least) that will fail because of this that are currently enabled.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -2,6 +2,7 @@
 //! This module defines error types and messages for SVD parsing and encoding
 
 pub use anyhow::{Context, Result};
+use core::u32;
 use once_cell::sync::Lazy;
 use regex::Regex;
 use xmltree::Element;
--- a/src/error.rs
+++ b/src/error.rs
@@ -74,6 +74,16 @@ pub enum NameError {
     Invalid(String, String),
 }
 
+#[derive(Clone, Debug, PartialEq, Eq, thiserror::Error)]
+pub enum ResetValueError {
+    #[error("Reset value 0x{0:x} doesn't fit in {1} bits")]
+    ValueTooLarge(u32, u32),
+    #[error("Reset value 0x{0:x} conflicts with mask '{1}'")]
+    MaskConflict(u32, u32),
+    #[error("Mask value 0x{0:x} doesn't fit in {1} bits")]
+    MaskTooLarge(u32, u32),
+}
+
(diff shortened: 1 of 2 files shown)


src/error.rs
```diff
@@ -96,3 +106,56 @@ pub(crate) fn check_name(name: &str, tag: &str) -> Result<()> {
         Err(NameError::Invalid(name.to_string(), tag.to_string()).into())
     }
 }
+
+pub(crate) fn check_reset_value(
+    size: &Option<u32>,
```
therealprof, line 111: `&Option<u32>`?

gkelly: That's how it's getting stored on RegisterProperties, so I figured I'd just take a reference. I can switch to non-ref if you think it's more ergonomic.

therealprof: I do think a reference to an `Option` of a primitive type  is a weird and unergonomic choice, yes. Not sure how much efforts it would be to change that but I certainly would prefer less ampersands in new code.

gkelly: Sounds good, just pushed with references switched to plain `Option<u32>`s.

burrbull:
Maybe it makes sense to add cargo feature something like `without-checks` to be possible to compile with disabled checks like `check_name`, this one and some other. This could be useful for tests.
cc @therealprof

gkelly:
> Maybe it makes sense to add cargo feature something like `without-checks` to be possible to compile with disabled checks like `check_name`, this one and some other. This could be useful for tests.
> cc @therealprof

I feel like `svd-parser` is already opinionated about correctness: name checking, enum variant validation. If `svd-parser` allows uncheck/unconstrained values to pass to consumers then it puts the burden on validation there, and so every consumer would just enable it...

burrbull:
Resolve conflicts, please.

gkelly:
> Resolve conflicts, please.

Done. Thank you for the reviews!

burrbull:
Add `use core::u32`

and fix `svd::registerproperties::tests::decode_encode` example.

gkelly:
> Add `use core::u32`

Done.

> and fix `svd::registerproperties::tests::decode_encode` example.

Done. Apologies, I wasn't building with `--features=untested`!

PR #124, opened 2020-06-02
Use the stable compiler to install `cross`

gkelly:
The `cross` tool is used during testing, but isn't a direct dependency.
Use the stable compiler when installing it instead of the active
toolchain during Travis tests to prevent having to update the MSRV to
support building `cross`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ rust:
   - nightly
   - stable
   # MSRV
-  - 1.35.0
+  - 1.36.0
 
 env: TARGET=x86_64-unknown-linux-gnu
 
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ This project is developed and maintained by the [Tools team][team].
 
 ## Minimum Supported Rust Version (MSRV)
 
-This crate is guaranteed to compile on stable Rust 1.35.0 and up. It *might*
+This crate is guaranteed to compile on stable Rust 1.36.0 and up. It *might*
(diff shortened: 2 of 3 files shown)


burrbull:
bors try

gkelly:
Okay, so now cross installs but the tool itself now seems to be trying to build something. Continuing looking...

burrbull:
1.36 still needed for crossbeam

therealprof:
Maybe we should just go GitHub Actions? At least we have a cached cross version to install there.

gkelly:
> 1.36 still needed for crossbeam

Ah, okay that's what I was misunderstanding.

gkelly:
I tested the equivalent process here:
```
$ rustup default 1.36.0
$ rustup install stable
$ cargo +stable install cross
$ cross version
cargo 1.36.0 (c4fcfb725 2019-05-15)
$ cross build --all-features
$ cross test --release --all-features
```

adamgreig:
bors try

gkelly:
@Emilgardis, you had a concern about the version bump in the other (closed) PR?

gkelly:
> 1.36 is fine but requires according documentation in the README.

I updated the `Minimum Supported Rust Version (MSRV)` heading (https://github.com/rust-embedded/svd/pull/124/files#diff-04c6e90faac2675aa89e2176d2eec7d8) in README.md, is there something else that should be added?

Emilgardis:
No concerns from me :)

PR #126, opened 2020-06-17
fix merge and add dim for field and register

tiannian:
1. fix dim merge for field and register.
2. add test for field and register.


burrbull:
cc @therealprof

therealprof (approved):
LGTM, too.

bors r+

PR #128, opened 2020-10-25
Add support for 64-bit addresses and values

Disasm:


Disasm:
Note that the additional changes I made are mostly "fix everything that doesn't compile anymore" and I could have missed something.

src/svd/peripheral.rs
```diff
@@ -76,7 +76,7 @@ pub struct Peripheral {
 #[derive(Clone, Debug, Default, PartialEq)]
 pub struct PeripheralBuilder {
```
burrbull, line 77: Maybe it makes sense to parametrize structures with size: `PeripheralBuilder<U>` instead of replacing all u32 with u64?

Disasm: Hmm, looks like unnecessary complication, given that the size type is not something the user can change.

Disasm:
Rebased to fix conflicts and added support for 64-bit writeConstraint and enum values.

burrbull (approved):

therealprof (approved):
LGTM, too. Thanks.

bors r+

therealprof:
Oh crap... This is still on Travis CI. 🙄

PR #129, opened 2020-10-25
Minor fixes for error-messages & fields with odd bitRange

Rahix:
Hello!

I'm working with yet another broken SVD file and found a few rough edges while trying to convert it.  This PR contains 3 small improvements to error display (proper errors instead of panicking)  and a fix for fields that are 32 bits wide _and_ contain enumerated values:

### 58ddcde48052632c8fddaeb9a453a3d7a70a9016 - Add proper error for missing array/cluster placeholder
Proper error message when an array or cluster name does not contain `%s`.  Previously, this would yield a cryptic panic message without any info about the offending register/field.

### 784eeb9a10fc30acaa6d2ed0cce05ed5eb67d6c4 - ResetValueError: Print mask as hexadecimal
Minor style tweak of this error.

### 016f6ba10fbdd0c3380f87c718c1a6c348e92314 - fieldinfo: Properly deal with 32-wide fields with enumerated values
The expression `2_u32.pow(self.bit_range.width)` will overflow when the bit_range has a width of 32.  As we know that a bit_range width of 32 will mean all enumerated values must be contained in the range anyway (as long as the `EnumeratedValue.value` field is `u32`), we can skip the checks and avoid dealing with this problem.

Another solution would be to switch to `RangeInclusive<u32>` but that yields a larger diff which I wanted to avoid.

### da17b29bc4bfc2fc734d4fca023cdc7d0fc7f4f5 - fieldinfo: Disallow bitRange width == 0
A field with width 0 does not actually contain any bits so this does not make sense.  Raise an error here to prevent later code assuming a bitRange width > 0 from panicking with cryptic messages.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Peripherals are processed in parallel with `rayon`
 - Serde now skips empty tags
 - Fix: bug in `addressUnitBits`
+- Fix: panic with 32 bit wide fields that have enumerated values
+- Fix: produce error on 0-width fields
+- Fix: error instead of panic when an array/cluster name is missing the `%s` placeholder
 
 ## [v0.9.0] - 2019-11-17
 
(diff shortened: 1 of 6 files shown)


therealprof (commented):
Can you add a CHANGELOG entry? Otherwise looks good to me.

Rahix:
Added entries to the CHANGELOG.

therealprof (approved):
LGTM, thank you!

bors r+

Issue #116, opened 2020-04-25
remove `unproven` flag


burrbull:
Encoding works fine as for me.

I think we should prove it and remove `unproven` option.

Alternative:
Split this crate on several like `svd-base`, `svd-parser`, `svd-encoder`, `svd-utils`, tests.

MarcoIeni:
> I think we should prove it and remove unproven option.

This means writing some tests, right? :)

In my project I added a test in which I test this library.
In the test I decode an xml, I encode it and then I decode it again.
The test asserts that the two decoded xml are equal.
Here is the test:
https://github.com/MarcoIeni/svdtools/blob/riir/tests/svd_parser.rs

Maybe we could start by writing a (integration?) test like this.

PR #131, opened 2020-12-26
Remove unproven

brainstorm:
Fixes https://github.com/rust-embedded/svd/issues/116

@MarcoIeni @burrbull Let's do what you suggest on https://github.com/rust-embedded/svd/issues/116, shall we? See the integration [test mostly borrowed from @MarcoIeni's `svd_parser`](https://github.com/rust-embedded/svd/pull/131/files#diff-eb771f54a1fdc1c330a82981d7d27d1202a5cbe3a59581ff67af5e3361f5f90e)

Getting the following error for now, ~seems like getting the same capitalization (and perhaps ordering?) might be challenging for integration testing purposes~... nope, it's just `<fields />` being empty,  I guess the SVD parser should tolerate that?:

```xml
    <peripheral>
           <name>RNG</name>
           <description>RNG register</description>
           <baseAddress>0x3FF20E44</baseAddress>
           <addressBlock>
                       <offset>0</offset>
                       <size>32</size>
                       <usage>RNG register</usage>
           </addressBlock>
           <registers>
                  <register>
                       <name>rng</name>
                       <fields />
                       <description>RNG register</description>
                       <addressOffset>0</addressOffset>
                       <size>32</size>
                       <access>read-only</access>
                      <resetValue>0</resetValue>
                   </register>
        </registers>
     </peripheral>
```

```rust
---- read_and_write stdout ----
thread 'read_and_write' panicked at 'Failed to parse the SVD file into Rust structs: In device `esp8266`

Caused by:
    0: In peripheral `RNG`
    1: In register `rng`
    2: Register have `fields` tag, but it is empty', tests/integration.rs:14:38
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```


brainstorm:
A couple of bits I've realized while thinking about this PR:

1) Ideally this integration testing should be run against https://github.com/posborne/cmsis-svd, to expose this parser to many ICs, even those not even supported by @rust-embedded, since they don't have HALs/PACs (yet)... if so, would it make sense to pull the repo via git submodules or other mechanisms are preferred? Perhaps this is already done somewhere else (in [svdtools](https://github.com/stm32-rs/svdtools/issues/9))?

2) I mentioned the introduction of a `lax` feature flag in https://github.com/rust-embedded/svd/pull/131/commits/e17cbae99667dcd9d681ecbb80313a64b02f477e ... I've seen the pros/cons mentioned in passing somewhere else, but it would be good to reference those discussions here and reach consensus to get it done for good?

tests/integration.rs
```diff
@@ -0,0 +1,54 @@
...
+    // Write (encode/serialize) SVD
+    let device_svd = svd::encode(&device).unwrap();
+
+    // Read and Write (deserialize) SVD again
+    let device_svd_deserialized = svd::encode(&svd::parse(&device_svd).unwrap()).unwrap();
+
+    assert_eq!(device_svd, device_svd_deserialized)
+}
+
+// fn diff_svds(orig: Path, dest: Path) -> Result<Diff> {
```
brainstorm, line 25: @MarcoIeni, Are you exploring https://github.com/MarcoIeni/delta for SVD diffing currently?

burrbull:
>     1. Ideally this integration testing should be run against https://github.com/posborne/cmsis-svd, to expose this parser to many ICs, even those not even supported by @rust-embedded, since they don't have HALs/PACs (yet)... if so, would it make sense to pull the repo via git submodules or other mechanisms are preferred? Perhaps this is already done somewhere else (in [svdtools](https://github.com/stm32-rs/svdtools/issues/9))? **EDIT:** CI seems to be thoroughly discussed on [stm32-rs/svdtools#6](https://github.com/stm32-rs/svdtools/issues/6)...  only for STM32 though.
> 

`cmsis-svd` is already connected via submodules. Or I don't understand question.

>     2. I mentioned the introduction of a `lax` (or `strict`) feature flag in [e17cbae](https://github.com/rust-embedded/svd/commit/e17cbae99667dcd9d681ecbb80313a64b02f477e) ... I've seen the pros/cons mentioned in passing somewhere else, but it would be good to reference those discussions here and reach consensus to get it done for good?

Yes, we need to add `strict` feature for several new checks.

brainstorm:
> 
> `cmsis-svd` is already connected via submodules. Or I don't understand question.


Ouch, my fault for not doing a `git clone --recursive` 🤦🏻  Thanks.


> > ```
> > 2. I mentioned the introduction of a `lax` (or `strict`) feature flag in [e17cbae](https://github.com/rust-embedded/svd/commit/e17cbae99667dcd9d681ecbb80313a64b02f477e) ... I've seen the pros/cons mentioned in passing somewhere else, but it would be good to reference those discussions here and reach consensus to get it done for good?
> > ```
> 
> Yes, we need to add `strict` feature for several new checks.
> CMSIS SVD specification says that some tags can absent, but can't be empty. So we should have such checks, but can hide them under feature.
> https://www.keil.com/pack/doc/CMSIS/SVD/html/elem_registers.html#elem_fields

Good, is that being worked on under some branch?

burrbull:
> Good, is that being worked on under some branch?

No, as I know.

You can see this also:
https://github.com/rust-embedded/svd2rust/pull/447

brainstorm:
Ah, good to see the integration tests there, I suspected this was not the right place to put those, thanks for the pointer. 

Shall I remove the integration test from here and you merge this PR with just the `unproven` flag removed or just close this PR?

tests/integration.rs
MarcoIeni: no, I did some work on delta just for fun, not svd related reasons :P

burrbull:
> Ah, good to see the integration tests there, I suspected this might not be the right place to put those, thanks for the pointer!
> 
> Shall I remove the integration test from here and you merge this PR with just the `unproven` flag removed or just close this PR and let you carry on? ;)

1 think 3 PRs should be here:
1) remove `unproven`
2) hide some checks under `strict`
3) add more tests

brainstorm:
Alright, backtracked on this PR so that it just fits point 1: remove unproven.

Cheers!

tests/bad_svd.rs
```diff
@@ -4,7 +4,7 @@ use svd_parser as svd;
 fn arm_sample_faulty() {
     let xml = include_str!(concat!(
         env!("CARGO_MANIFEST_DIR"),
-        "/tests/ARM_Sample_faulty.svd"
+        "/tests/data/ARM_Sample_faulty.svd"
```
MarcoIeni, line 7: you missed this I think

.gitignore
```diff
@@ -2,4 +2,3 @@
 [._]*.sw[a-p]
 Cargo.lock
 target
-tests/
```
MarcoIeni, line 5: is there a reason why `tests` directory was ignored?

brainstorm: 🤷🏻‍♂️ ... I guess that @burrbull was generating some files automatically and did not want to version them in the past? Can't see any reason to keep ignoring it TBH.

tests/bad_svd.rs
brainstorm: You mean reverting it as it was? I think that this structure is more intuitive (keep tests and data separate)... it's a minimal change and tests keep passing as they were anyway?:

```rust
$ cargo test
(...)
running 1 test
test arm_sample_faulty ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

MarcoIeni: Sorry, I haven't seen you moved the file!

.gitignore
burrbull: It is not my repo initially :)

burrbull:
LGTM. Squash commits and modify changelog please.

MarcoIeni:
> Squash commits and modify changelog please.

Just out of curiosity. Why should he squash commits? You can select the option "Squash and merge" from github.

burrbull:
We use bors for tests and merging as I remember.

MarcoIeni:
you are right, thanks!

burrbull (approved):

burrbull:
Looks like `ci` is broken. We need to move it on Github Actions and update `cross`

MarcoIeni:
> Looks like `ci` is broken. We need to move it on Github Actions and update `cross`

See #132
I can do it!

ForsakenHarmony:
> We use bors for tests and merging as I remember.

You can configure bors to squash merge

MarcoIeni:
> You can configure bors to squash merge

Yes, but I don't remember where, therealprof told me that sometimes they prefer to keep commit history, so they don't want to squash and merge everytime.

PR #133, opened 2020-12-26
switch to GH actions

### LINKED ISSUE #132: Is Continuous Integration working?
I was looking at #131 and I don't see something that checks if the code is formatted and that tests are working on each commit you do in the PR.

Is travis still working?
We add some issues with it in [delta](https://github.com/dandavison/delta/issues/396), might be related?
If you want I can add GitHub actions in this project really easily.

MarcoIeni:
Closes #132

Questions:
- why is cross needed?
- Do you think I covered everything Travis was doing?


MarcoIeni:
I don't get why actions are not working, even on my branch: https://github.com/MarcoIeni/svd/actions

burrbull:
bors try

brainstorm:
One thing that `cross` offers is (some) QEMU MCU emulation... I **guess** that using cross was a way to future-proof CI so that very simple emulation tests could be implemented in the future?

Other than that, LGTM :+1+

MarcoIeni:
I think we have to change this:

https://github.com/rust-embedded/svd/blob/051e0b26cc439a8bf002c61eb47da111a6adab15/bors.toml#L2

MarcoIeni:
> I **guess** that using cross was a way to future-proof CI so that very simple emulation tests could be implemented in the future?

If we are not using it at the moment we can remove the `use-cross` flag. We can add it in the future if needed.

burrbull:
cc @therealprof

therealprof (changes requested):
Please follow the conventions outlined in https://github.com/rust-embedded/wg/blob/master/ops/post-transfer.md

Specifically the trigger rules and bors.toml are not sufficient yet.

MarcoIeni:
> I think we have to change this:
> 
> https://github.com/rust-embedded/svd/blob/051e0b26cc439a8bf002c61eb47da111a6adab15/bors.toml#L2

I completely removed the `status` field of bors, because maybe it is not required with github actions. Is this right?
Can you check if this field is present in some other project that uses bors and github actions?

burrbull:
@therealprof can you check bors settings/permissions for this repo?

burrbull (dismissed):

therealprof:
> @therealprof can you check bors settings/permissions for this repo?

Can do but I think it should work already just fine.

.github/workflows/ci.yml
```diff
@@ -12,14 +12,18 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        rust_version: [1.36.0, stable, nightly]
+        rust: [ 1.36.0, stable ]
+        include:
+          # Test nightly but don't fail the build.
+          - rust: nightly
+            experimental: true
```
adamgreig, line 19: By the way, `experimental` by itself doesn't do or mean anything; you need to set `continue-on-error` based on the value of `experimental` (see the wg template CI for an example)

MarcoIeni: Thank you, fixed!

MarcoIeni:
Should I remove `"S-waiting-on-team"` from `bors.toml`? Because it is not present in the template.

therealprof:
> Should I remove `"S-waiting-on-team"` from `bors.toml`? Because it is not present in the template.

Don't bother. Doesn't do any harm.

therealprof:
Mind squashing your commits?

MarcoIeni:
I will do it.
For the future, have you considered adding `use_squash_merge = true` boolean variable to bors.toml? See https://bors.tech/documentation/

burrbull (approved):

therealprof:
Having separate commits can be very useful so auto-squashing is not a good idea.

therealprof (approved):
LGTM, thanks.

bors r+

Issue #109, opened 2020-04-03
Change xml crate


MarcoIeni:
I am working on [this](https://github.com/stm32-rs/svdtools/issues/9) issue and I noticed that this library is really slow if compared to what I built by using [quick-xml](https://github.com/tafia/quick-xml).

I am assuming that the difference of performances is due to the different xml library used.

If you want to compare the two solutions:
```
git clone -b riir https://github.com/MarcoIeni/svdtools
cd svdtools

# try version the version with quick-xml 
git checkout be257c679028d10c6d86e0ae9c869a0ee421cb7c
cargo build --release
time ./target/release/svdtools mmap example/stm32l4x2.svd

# try the version that uses svd-parser
git checkout fad7abfd923312ec95f307286a05017be1218217
cargo build --release
time ./target/release/svdtools mmap example/stm32l4x2.svd
```

These are the results I get on my laptop by running the executable about ten times:
```
quick-xml version:
minimum: 0.03s user 0.02s system 79% cpu 0.066 total
maximum: 0.08s user 0.01s system 87% cpu 0.100 total

svd-parser version:
minimum: 0.26s user 0.02s system 96% cpu 0.291 total
maximum: 0.30s user 0.02s system 95% cpu 0.332 total
```

Of course the version I built by using quick-xml is incomplete and doesn't handle errors well (I originally wrote it because I didn't know this crate existed), but I just wanted to tell you that I saw this huge performance difference.

Of course I know that these benchmark are not rigorous, but I think that they are sufficient at least to raise this point to your attention.

burrbull:
`svd-parser` now uses `roxmltree` which even faster
`svd-encoder` still uses `xmltree`

PR #138, opened 2021-01-31
Minicom XML parser/writer

burrbull:
I tried to replace `xmltree` with `minicom`, which uses `quick-xml`.
But I have problem. Since `0.13` version it requires `xmlns` namespace to be used to build each `Element`.

In other hand compiling with `0.12` version fails with:
```
error[E0277]: `Rc<minidom::namespace_set::NamespaceSet>` cannot be sent between threads safely
  --> src/types.rs:96:93
   |
96 |                     return Err(SVDError::InvalidBooleanValue(tree.clone(), text.clone(), e).into())
   |                                                                                             ^^^^ `Rc<minidom::namespace_set::NamespaceSet>` cannot be sent between threads safely
```

I like also `roxmltree` crate for high-level interaction with xml. But it can be used for parse only.

#109

### DIFF
--- a/src/svd/cluster.rs
+++ b/src/svd/cluster.rs
@@ -1,5 +1,5 @@
 use core::ops::Deref;
-use xmltree::Element;
+use minidom::Element;
 
 use crate::types::Parse;
 
@@ -31,11 +31,11 @@ impl Parse for Cluster {
     type Error = anyhow::Error;
 
     fn parse(tree: &Element) -> Result<Self> {
-        assert_eq!(tree.name, "cluster");
+        assert_eq!(tree.name(), "cluster");
 
         let info = ClusterInfo::parse(tree)?;
 
-        if tree.get_child("dimIncrement").is_some() {
+        if tree.get_child("dimIncrement", "").is_some() {
             let array_info = DimElement::parse(tree)?;
             check_has_placeholder(&info.name, "cluster")?;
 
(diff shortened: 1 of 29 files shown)


src/svd/cluster.rs
```diff
@@ -31,11 +32,11 @@ impl Parse for Cluster {
     type Error = anyhow::Error;
 
     fn parse(tree: &Element) -> Result<Self> {
-        assert_eq!(tree.name, "cluster");
+        assert_eq!(tree.name(), "cluster");
 
         let info = ClusterInfo::parse(tree)?;
 
-        if tree.get_child("dimIncrement").is_some() {
+        if tree.get_child("dimIncrement", NS).is_some() {
```
Emilgardis, line 39: I'd rather these be `crate::NS` instead of importing them

src/svd/registerproperties.rs
```diff
@@ -83,33 +83,36 @@ impl EncodeChildren for RegisterProperties {
...
     #[test]
     fn decode_encode() {
-        let example = String::from(
-            "
-            <mock>
+        let example = "
+            <mock xmlns=\""
+            .to_string()
+            + NS
+            + "\">
```
Emilgardis, line 94: This is unfortunate

brainstorm:
I suspect that with YaSerDe you might have better luck doing this? I had a bad time trying to tell `quick-xml` to basically differentiate between `attribute` and `child` but it was comparatively very easy with YaSerDe.

See the docs I just pushed there for reference: https://github.com/media-io/yaserde/pull/106

PR #144, opened 2021-04-26
RegIter

burrbull:
r? @therealprof


therealprof (dismissed):
I wonder where you're going with this but an iterator interface certainly makes sense... 😅

bors r+

therealprof (approved):
LGTM

bors r+

PR #149, opened 2021-04-28
Register properties everywhere

burrbull:


burrbull:
cc @therealprof

burrbull:
cc @Emilgardis 
I did not do this earlier because this is breaking change. But this should be done.

src/svd/registerinfo.rs
```diff
@@ -414,7 +377,7 @@ mod tests {
                 <addressOffset>0x8</addressOffset>
                 <alternateGroup>alternate_group</alternateGroup>
                 <alternateRegister>alternate_register</alternateRegister>
-                <size>32</size>
+                <size>0x20</size>
```
Emilgardis, line 380: ?

burrbull: reverted

burrbull: cc @Emilgardis

PR #157, opened 2021-10-29
WIP: Fail if a register array is derived

robamu:
I don't think this feature is implemented yet. I'm not sure how easy or difficult it is to be implemented, but this PR improved the error from something like 

```sh
[ERROR svd2rust] expected `,`
```

to something like

```sh
[ERROR svd2rust] In device `va108xx`
    
    Caused by:
        0: In peripheral `IRQSEL`
        1: Parsing register `PORTA[%s]` at 503:9
        2: No support for registers arrays which are derived
```

when parsing register blocks like this

```xml
        <register derivedFrom="INT_RAM_SBE">
          <dim>32</dim>
          <dimIncrement>4</dimIncrement>
          <name>PORTB[%s]</name>
          <description>PORTB Interrupt Redirect Selection</description>
          <addressOffset>0x080</addressOffset>
        </register>
```

If implementing the support for the  `derivedFrom` attribute on register arrays is not difficult (or I am indeed not doing something wrong) this PR could of course also be skipped. I also have not tested it with any other files other than the one I am trying to parse here: https://github.com/robamu/va108xx-rs

### DIFF
--- a/svd-parser/src/lib.rs
+++ b/svd-parser/src/lib.rs
@@ -190,7 +190,7 @@ pub enum SVDError {
     UnknownWriteConstraint,
     #[error("Multiple wc found")]
     MoreThanOneWriteConstraint,
-    #[error("No support for registers arrays which are derived")]
+    #[error("No support for register arrays which are derived")]
     DerivedRegisterArray,
     #[error("Unknown usage variant")]
     UnknownUsageVariant,
--- a/svd-parser/src/lib.rs
+++ b/svd-parser/src/lib.rs
@@ -190,8 +190,8 @@ pub enum SVDError {
     UnknownWriteConstraint,
     #[error("Multiple wc found")]
     MoreThanOneWriteConstraint,
-    #[error("No support for registers with arrays which are derived")]
-    DerivedRegisterWithArray,
+    #[error("No support for registers arrays which are derived")]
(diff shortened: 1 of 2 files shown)


robamu:
Alright, the parser function works without throwing the error and the error comes from the `convert_svd_register` function in the https://github.com/rust-embedded/svd2rust/blob/master/src/generate/peripheral.rs file , the `name_to_wrapped_ty` to be more precise. The syn error is propagated and maybe it would be a good idea to emitt additional information on a syn error as a first step.

burrbull:
Yea, svd2rust error system is outdated and should be reworked.

robamu:
I like how `svd-parse`  emits some additional information. Do you think something like that would be useful for the device generator in `svd2rust`  as well?

burrbull:
> I like how `svd-parse` emits some additional information. Do you think something like that would be useful for the device generator in `svd2rust` as well?

It would be excellent, but I don't have enough time for this now.
Saying about this particular issue, `svd2rust` work with already parsed `Device` structure, so you'll not be able to show line where svd2rust-specific error is occurred, only tree path.

robamu:
Can you have a look at https://github.com/rust-embedded/svd2rust/pull/544 ? 
I think this PR can be closed then.

PR #158, opened 2021-11-02
impl DeriveFrom for Register/Cluster/Field

burrbull:
~~Alternate is #159~~


burrbull:
cc @Emilgardis

Emilgardis (approved):
bors r+

PR #161, opened 2021-11-06
peripheral array

### LINKED ISSUE #41: Should Peripheral support array?
`Register` can take the variant `Single(RegisterInfo)` or `Array(RegisterInfo, RegisterArrayInfo)`.  The svd dim stuff is also valid for a peripheral, shouldn't that mean Peripheral should do the same?

What I suggest is to:
 1. Rename `RegisterArrayInfo` to `DimInfo` (or something else that's generic over these types, like `ArrayInfo`)
2. Rename `Peripheral` to `PeripheralInfo` (to make it analog to `RegisterInfo`)
3. Introduce the new `enum Peripheral` which is analog to `enum Register` 
4. Add `Peripheral::parse` analog to how `Register::parse` works

## Breakage
There are 3 crates listed on crates.io as dependent on this crate. Only [svd2rust](https://github.com/japaric/svd2rust) have received updates in 2017.

Right now svd2rust silently generates garbage code when given peripheral arrays. It should be trivial to apply a quick patch which crashes the program when faced peripheral arrays and works as previous when faced single peripherals. This patch is also a prerequisite of supporting peripheral arrays in svd2rust. In other words, the breakage caused by this patch would be trivial to fix and benefit svd2rust greatly both immediately and in the future.

burrbull:
Closes #41

### DIFF
--- a/svd-encoder/src/lib.rs
+++ b/svd-encoder/src/lib.rs
@@ -78,6 +78,7 @@ mod fieldinfo;
 mod interrupt;
 mod modifiedwritevalues;
 mod peripheral;
+mod peripheralinfo;
 mod register;
 mod registercluster;
 mod registerinfo;
--- a/svd-rs/src/peripheralinfo.rs
+++ b/svd-rs/src/peripheralinfo.rs
@@ -1,0 +1,304 @@
+use super::{
+    register::{RegIter, RegIterMut},
+    AddressBlock, BuildError, EmptyToNone, Interrupt, RegisterCluster, RegisterProperties,
+    SvdError, ValidateLevel,
+};
+
+/// Errors from [Peripheral::validate]
+#[derive(Clone, Debug, PartialEq, Eq, thiserror::Error)]
+pub enum Error {
+    /// The peripheral has no registers, but specified a `<registers>` tag.
+    #[error("Peripheral have `registers` tag, but it is empty")]
+    EmptyRegisters,
+}
+
+/// A description of a peripheral in the [device](crate::Device), describing, for example, the [memory mappings](crate::RegisterInfo).
+#[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
+#[cfg_attr(feature = "serde", serde(rename_all = "camelCase"))]
+#[derive(Clone, Debug, PartialEq)]
+#[non_exhaustive]
+pub struct PeripheralInfo {
+    /// The string identifies the peripheral. Peripheral names are required to be unique for a device
+    pub name: String,
+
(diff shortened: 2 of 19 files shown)


svd-encoder/src/lib.rs
```diff
@@ -78,6 +78,7 @@ mod fieldinfo;
 mod interrupt;
 mod modifiedwritevalues;
 mod peripheral;
+mod peripheralinfo;
```
Emilgardis, line 81: imo, this should be `peripheral_info` but since we're already using this format, it makes sense.

svd-rs/src/peripheralinfo.rs
```diff
@@ -0,0 +1,304 @@
...
+        if !lvl.is_disabled() {
+            self.validate(lvl)
+        } else {
+            Ok(())
+        }
+    }
+
+    /// Validate the [`Peripheral`]
+    pub fn validate(&mut self, lvl: ValidateLevel) -> Result<(), SvdError> {
+        // TODO
```
Emilgardis, line 263: what is this todo? Comes from https://github.com/rust-embedded/svd/blob/81459691f379811650e7be81d81043e06b57437d/svd-rs/src/peripheral.rs#L263
But I think it was left in by mistake

svd-rs/src/peripheralinfo.rs
```diff
@@ -0,0 +1,304 @@
...
+#[derive(Clone, Debug, PartialEq, Eq, thiserror::Error)]
+pub enum Error {
+    /// The peripheral has no registers, but specified a `<registers>` tag.
+    #[error("Peripheral have `registers` tag, but it is empty")]
+    EmptyRegisters,
+}
+
+/// A description of a peripheral in the [device](crate::Device), describing, for example, the [memory mappings](crate::RegisterInfo).
+#[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
+#[cfg_attr(feature = "serde", serde(rename_all = "camelCase"))]
```
Emilgardis, line 17: this could be
```suggestion
#[cfg_attr(
        feature = "serde",
        derive(serde::Deserialize, serde::Serialize),
        serde(rename_all = "camelCase")
)]
```
but I have no real opinion on it

Emilgardis (approved):
lgtm, some minor things, not a blocker

svd-rs/src/peripheralinfo.rs
burrbull: I'd change this in another PR.

burrbull:
bors r=Emilgardis

PR #163, opened 2021-11-07
Read action

burrbull:
depends on #162


burrbull:
rebased
cc @Emilgardis

Emilgardis (approved):
Awesome!

Should probably have a look at https://github.com/rust-embedded/svd/wiki/Compatibility-with-SVD-spec and update it to current features

bors r+

PR #165, opened 2021-11-07
array constructors

burrbull:

Issue #164, opened 2021-11-07
Proposal: struct instead of enum for arrays


burrbull:
Instead of enum:
```rust
pub enum Register {
    /// A single register.
    Single(RegisterInfo),
    /// An array of registers.
    Array(RegisterInfo, DimElement),
}
```
I propose to use struct:
```rust
struct Register {
    info: RegisterInfo,
    dim: Option<DimElement>,
}
```
where `dim: None` corresponds to `Register::Single`.
The main goal is to simplify serialize/deserialize.

cc @Emilgardis  @therealprof  @adamgreig

Emilgardis:
It would also be possible to do this with

```rust
#[cfg_attr(
    feature = "serde",
    derive(serde::Serialize, serde::Deserialize),
    serde(untagged)
)]
pub enum Register {
    /// A single register.
    Single {
        /// Register information
        info: RegisterInfo,
    },
    /// An array of registers.
    Array {
        /// Register information
        info: RegisterInfo,
        /// Array dim
        dim: DimElement,
    },
}
```

I wonder which one is the best

burrbull:
> It would also be possible to do this with
> 
> ```rust
> #[cfg_attr(
>     feature = "serde",
>     derive(serde::Serialize, serde::Deserialize),
>     serde(untagged)
> )]
> pub enum Register {
>     /// A single register.
>     Single {
>         /// Register information
>         info: RegisterInfo,
>     },
>     /// An array of registers.
>     Array {
>         /// Register information
>         info: RegisterInfo,
>         /// Array dim
>         dim: DimElement,
>     },
> }
> ```
> 
> I wonder which one is the best

Unfortunately the result is different from I expect.
![изображение](https://user-images.githubusercontent.com/3072754/141054853-fed92a16-903f-42b9-ad11-85dd836fb653.png)

burrbull:
I can take identical result with:
```rust
#[derive(Clone, Debug, PartialEq)]
#[cfg_attr(
    feature = "serde",
    derive(serde::Serialize, serde::Deserialize),
    serde(untagged)
)]
pub enum Register {
    /// A single register.
    Single(RegisterInfo),
    /// An array of registers.
    Array(RegisterArray),
}

#[derive(Clone, Debug, PartialEq)]
#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
/// ...
pub struct RegisterArray {
    #[cfg_attr(feature = "serde", serde(flatten))]
[...]

PR #166, opened 2021-11-07
Array struct

burrbull:
implement #164 
depends on #165


Emilgardis:
why split up the changes into different prs?

burrbull:
First PR (#165) is not breaking. Just add some sugar.

PR #172, opened 2021-11-09
name + bitrange fix

Emilgardis:
- check for wrong range size
- add function to get name of cluster


Emilgardis:
the name() fn can probably be discarded.

burrbull:
> the name() fn can probably be discarded.

as we have `Deref`, `&self.name` should already work

PR #173, opened 2021-11-10
common serde helpers for arrays

burrbull:


burrbull:
cc @Emilgardis

svd-rs/src/lib.rs
```diff
@@ -238,3 +238,21 @@ impl<T> EmptyToNone for Option<Vec<T>> {
         self.and_then(|v| if v.is_empty() { None } else { Some(v) })
     }
 }
+
+#[cfg(feature = "serde")]
+#[derive(serde::Serialize)]
+struct SerArray<'a, T> {
+    #[cfg_attr(feature = "serde", serde(flatten))]
```
Emilgardis, line 245: These cfg_attrs are useless, since the struct is only output with the serde feature enabled

Emilgardis (changes requested):

svd-rs/src/lib.rs
burrbull: good catch

burrbull: fixed

Emilgardis (approved):
bors r+

PR #181, opened 2021-12-09
use generic enum for arrays, remove OptIter, add more methods

burrbull:


burrbull:
cc @Emilgardis

svd-rs/src/array.rs
```diff
@@ -0,0 +1,122 @@
...
+where
+    T: Name,
+{
+    /// Returns list of single or array names
+    pub fn names(&self) -> Vec<Cow<str>> {
+        match self {
+            Self::Single(info) => vec![info.name().into()],
+            Self::Array(info, dim) => dim
+                .indexes()
+                .map(|i| info.name().replace("[%s]", &i).replace("%s", &i).into())
```
Emilgardis, line 64: Maybe this should be a generic Fn to plug in a choosen function to map into Cow<str> ?

svd-rs/CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+- Use generic enum for types, which can be collected into SVD arrays
```
Emilgardis, line 10: Unnecessary comma

svd-rs/src/enumeratedvalues.rs
```diff
@@ -155,4 +155,14 @@ impl EnumeratedValues {
     pub fn usage(&self) -> Usage {
         self.usage.unwrap_or_default()
     }
+
+    /// Get enumeratedValue by name
```
Emilgardis, line 159:
```suggestion
    /// Get `enumeratedValue` by name
```

Emilgardis (changes requested):

svd-rs/src/array.rs
burrbull: Sorry, but I don't understand what you means.

svd-rs/CHANGELOG.md
thejpster: The comma tells me that it is the "generic enum" that can be collected into SVD arrays, as opposed to "types". I don't know which is correct at first glance.

svd-rs/src/array.rs
burrbull: @Emilgardis I've deleted additional array functions for now. I'll make another PR for this

burrbull:
cc @Emilgardis
Look again, please.

svd-rs/src/array.rs
```diff
@@ -0,0 +1,105 @@
+use super::{DimElement, Name};
+use core::ops::{Deref, DerefMut};
+
+/// A single SVD instance or array of instances
+#[derive(Clone, Debug, PartialEq)]
+pub enum SingleArray<T> {
```
Emilgardis, line 6: I think the name `SingleArray` is a bit confusing. Not sure what alternatives there are though.

Emilgardis (approved):
LGTM, would personally want another name for the array enum but not sure what

svd-rs/src/array.rs
burrbull: I'm also not fun on this. I named this `SingleArray` because I thought about changing this enum in future to:
```rust
pub enum SingleArray<T> {
    /// A single instance
    Single(T),
    /// An array of instances
    Array(Array<T>),
}
```
where
```rust
pub struct Array<T> {
    dim: DimElement,
    info: T
}
```
This would let us add some useful methods on `Array`. But this is breaking change, especially in `match` expessions.

burrbull: `MaybeArray`

burrbull:
bors r=Emilgardis

PR #182, opened 2021-12-29
merge register with registerinfo, same with other info

burrbull:
As `register` now contain only type alias, no sense to have 2 separate files for Register and RegisterInfo


burrbull:
cc @Emilgardis

Emilgardis (approved):
bors r+

PR #187, opened 2022-01-02
v0.13

burrbull:


thejpster:
Hmm. Looks OK, but I think I'd prefer to see the new example and the serialisation changes in a separate PR.

burrbull:
> Hmm. Looks OK, but I think I'd prefer to see the new example and the serialisation changes in a separate PR.

#188

burrbull:
cc @rust-embedded/tools

thejpster (approved):

burrbull:
Need to merge #192 first

adamgreig (approved):
LGTM. I noticed we don't have links in the changelog any more, but still use the `[v0.13.0]` style headers. Maybe worth adding links? They would have to be to the whole-repository but I guess it's OK.

burrbull:
> LGTM. I noticed we don't have links in the changelog any more, but still use the `[v0.13.0]` style headers. Maybe worth adding links? They would have to be to the whole-repository but I guess it's OK.

Ok. I'll try to do this

adamgreig (approved):
Thanks!

bors merge

PR #195, opened 2022-02-12
encode dimIndex ranges

burrbull:
cc @adamgreig 
Partially related to https://github.com/stm32-rs/stm32-rs/pull/701#issuecomment-1036982033

### DIFF
--- a/svd-encoder/CHANGELOG.md
+++ b/svd-encoder/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+- Encode "dimIndex" ranges
+
 ## [v0.13.0] - 2022-01-02
 
 - Bump `svd-rs`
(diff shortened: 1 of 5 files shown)


burrbull:
cc @Emilgardis

adamgreig (approved):
LGTM. Maybe worth fixing the clippy lint about the `return Some(min..=max);`?

I also think the `max.wrapping_sub(min)` could just be `(max - min)` since it can't ever underflow, but it doesn't really matter here.

bors d+

adamgreig (approved):
hah, what use is bors delegating if as soon as you push a commit it revokes approval?

bors r+

therealprof:
> hah, what use is bors delegating if as soon as you push a commit it revokes approval?
> 
> bors r+

Yeah, need to change the configuration to not dismiss a review after a new commit.

Emilgardis:
> > hah, what use is bors delegating if as soon as you push a commit it revokes approval?
> > bors r+
> 
> Yeah, need to change the configuration to not dismiss a review after a new commit.

it's quite easy to change branch protection rules, should we?

image

therealprof:
I'm all for it, especially for projects with little risk of abuse (like sneaking in changes after approval of unmergeable PRs). However we'd have to be more permissive in our rules: https://github.com/rust-embedded/wg/blob/master/ops/post-transfer.md#repository-settings

PR #198, opened 2022-04-15
Expand & ignore_enums parse options

burrbull:
cc @noppej

### DIFF
--- a/svd-parser/src/expand.rs
+++ b/svd-parser/src/expand.rs
@@ -1,0 +1,400 @@
+//! Provides [expand] method to convert arrays, clusters and derived items in regular instances
+
+use anyhow::{anyhow, Result};
+use std::collections::HashMap;
+use std::mem::take;
+use svd_rs::{
+    array::names, cluster, field, peripheral, register, BitRange, Cluster, ClusterInfo, DeriveFrom,
+    Device, EnumeratedValues, Field, Peripheral, Register, RegisterCluster,
+};
+
+#[derive(Clone, Debug, Default)]
+struct Index<'a> {
+    peripherals: HashMap<String, &'a Peripheral>,
+    clusters: HashMap<String, &'a Cluster>,
+    registers: HashMap<String, &'a Register>,
+    fields: HashMap<String, &'a Field>,
+    evs: HashMap<String, &'a EnumeratedValues>,
+}
+
+impl<'a> Index<'a> {
+    fn add_peripheral(&mut self, p: &'a Peripheral) {
+        let path = &p.name;
+        if let Peripheral::Array(info, dim) = p {
+            for name in names(info, dim) {
+                let path = name;
+                for r in p.registers() {
+                    self.add_register(&path, r);
+                }
+                for c in p.clusters() {
+                    self.add_cluster(&path, c);
+                }
+                self.peripherals.insert(path, p);
+            }
+        }
+        for r in p.registers() {
+            self.add_register(path, r);
+        }
+        for c in p.clusters() {
+            self.add_cluster(path, c);
+        }
+        self.peripherals.insert(path.into(), p);
+    }
+
+    fn add_cluster(&mut self, path: &str, c: &'a Cluster) {
+        if let Cluster::Array(info, dim) = c {
+            for name in names(info, dim) {
+                let cpath = format!("{}.{}", path, name);
+                for r in c.registers() {
+                    self.add_register(&cpath, r);
+                }
+                for c in c.clusters() {
+                    self.add_cluster(&cpath, c);
+                }
+                self.clusters.insert(cpath, c);
+            }
+        }
+        let cpath = format!("{}.{}", path, c.name);
+        for r in c.registers() {
+            self.add_register(&cpath, r);
+        }
+        for c in c.clusters() {
+            self.add_cluster(&cpath, c);
+        }
+        self.clusters.insert(cpath, c);
+    }
+    fn add_register(&mut self, path: &str, r: &'a Register) {
+        if let Register::Array(info, dim) = r {
+            for name in names(info, dim) {
+                let rpath = format!("{}.{}", path, name);
+                for f in r.fields() {
+                    self.add_field(&rpath, f);
+                }
+                self.registers.insert(rpath, r);
+            }
+        }
+        let rpath = format!("{}.{}", path, r.name);
+        for f in r.fields() {
+            self.add_field(&rpath, f);
+        }
+        self.registers.insert(rpath, r);
+    }
+    fn add_field(&mut self, path: &str, f: &'a Field) {
+        if let Field::Array(info, dim) = f {
+            for name in names(info, dim) {
+                let fpath = format!("{}.{}", path, name);
+                for evs in &f.enumerated_values {
+                    if let Some(name) = evs.name.as_ref() {
+                        let epath = format!("{}.{}", fpath, name);
+                        self.evs.insert(epath, evs);
+                    }
+                }
+                self.fields.insert(fpath, f);
+            }
+        }
+        let fpath = format!("{}.{}", path, f.name);
+        for evs in &f.enumerated_values {
+            if let Some(name) = evs.name.as_ref() {
+                let epath = format!("{}.{}", fpath, name);
+                self.evs.insert(epath, evs);
+            }
+        }
+        self.fields.insert(fpath, f);
+    }
+
+    pub fn create(device: &'a Device) -> Self {
+        let mut index = Self::default();
+        for p in &device.peripherals {
+            index.add_peripheral(p);
+        }
+        index
+    }
+
+    pub fn get_base_peripheral(&self, path: &str) -> Option<&Peripheral> {
+        self.peripherals
+            .get(path)
+            .and_then(|&p| match &p.derived_from {
+                None => Some(p),
+                Some(dp) => self.get_base_peripheral(dp),
+            })
+    }
+}
+
+fn expand_register_cluster(
+    regs: &mut Vec<RegisterCluster>,
+    rc: RegisterCluster,
+    path: &str,
+    index: &Index,
+) -> Result<()> {
+    match rc {
+        RegisterCluster::Cluster(c) => expand_cluster_array(regs, c, path, index)?,
+        RegisterCluster::Register(r) => expand_register_array(regs, r, path, index)?,
+    }
+    Ok(())
+}
+
+fn expand_cluster_array(
+    regs: &mut Vec<RegisterCluster>,
+    mut c: Cluster,
+    path: &str,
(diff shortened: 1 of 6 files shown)


burrbull:
1. Where should `expand` be? In `svd-rs`, `svd-parser` or external crate? `svd-parser` works with XML only (fast + advanced error checks). YAML and JSON use `serde` (directly on `svd-rs`).
2. Should it be under feature?
3. Need more tests.

adamgreig:
What does expand do?

burrbull:
> What does expand do?

It makes copy of all "derived" peripherals, registers, etc. same time as expanding of all arrays.

noppej:
> 3\. Need more tests.

I can help with testing for use in `probe-rs-debugger`. I have SVD files with 'derived' peripherals and registers, but none of my SVD's have arrays. Can you point me at some that have arrays?

noppej:
> 2\. Should it be under feature?

It is hard for me to imagine a use case where having derived peripherals and expanded arrays is not an absolute requirement. If it was up to me, I'd say, no need for a feature.

> It makes copy of all "derived" peripherals, registers, etc. same time as expanding of all arrays.

This is a really useful addition, and will obsolete a lot of ugly code on my side :).   Thank you!

burrbull:
> Can you point me at some that have arrays?

Almost any patched SVD from https://stm32-rs.github.io/stm32-rs/

burrbull:
> It is hard for me to imagine a use case where having derived peripherals and expanded arrays is not an absolute requirement.

Initially `svd-parser` was a part of `svd2rust` (generate pac's from SVD). It is much more better when `svd2rust` generates 1 Rust type for identical registers/peripherals. So our main goal was to `collect` registers in arrays/clusters instead of expanding arrays in independent registers.

noppej:
> Almost any patched SVD from https://stm32-rs.github.io/stm32-rs/

The only ST device I have for testing is an `STM32H757_cm7`. When I try to retrieve [its svd](https://stm32-rs.github.io/stm32-rs/stm32h757_cm7.svd.patched) from that site, I get a 404. Is that known/expected or should I create an issue in that repo?

adamgreig:
Oops, that's an error, I guess the filename changed and the index wasn't updated. Could you open an issue please?

burrbull:
4 month old
https://raw.githubusercontent.com/tinygo-org/stm32-svd/main/svd/stm32h747cm7.svd

noppej:
> https://stm32-rs.github.io/stm32-rs/stm32h747cm7.svd.patched

Thanks. How would you like me to report back results of testing?

burrbull:
> Thanks. How would you like me to report back results of testing?

With thumb emoji :) 

Whichever way you feel comfortable

noppej:
I've implemented this in a [draft PR](https://github.com/probe-rs/probe-rs/pull/1090) for `probe-rs-debugger`. It significantly reduced the amount of code required on probe-rs side, added the ability to expand arrays and clusters (definitely valuable) and also improved performance of this part of the code in the debugger.   Thanks for an awesome crate!

.gitmodules
```diff
@@ -1,3 +0,0 @@
-[submodule "cmsis-svd"]
-	path = cmsis-svd
-	url = git://github.com/posborne/cmsis-svd.git
```
Emilgardis, line 3: why was this submodule removed?

svd-parser/src/expand.rs
```diff
@@ -0,0 +1,396 @@
...
+) -> Result<()> {
+    let rpath = if let Some(dpath) = r.derived_from.as_ref() {
+        let rpath = dpath.to_string();
+        if let Some(d) = index
+            .registers
+            .get(dpath)
+            .or_else(|| index.registers.get(&format!("{}.{}", path, dpath)))
+        {
+            if d.derived_from.is_some() {
+                return Err(anyhow!("Multiple derive for {} is not supported", dpath));
```
Emilgardis, line 210:
```suggestion
                return Err(anyhow!("multiple derive for {} is not supported", dpath));
```

svd-parser/src/expand.rs
```diff
@@ -0,0 +1,396 @@
...
+            }
+            r = r.derive_from(d);
+            r.derived_from = None;
+        } else {
+            return Err(anyhow!("Register {} not found", dpath));
```
Emilgardis, line 215:
```suggestion
            return Err(anyhow!("register {} not found", dpath));
```

svd-parser/src/expand.rs
```diff
@@ -0,0 +1,396 @@
...
+) -> Result<()> {
+    let fpath = if let Some(dpath) = f.derived_from.as_ref() {
+        let fpath = dpath.to_string();
+        if let Some(d) = index
+            .fields
+            .get(dpath)
+            .or_else(|| index.fields.get(&format!("{}.{}", rpath, dpath)))
+        {
+            if d.derived_from.is_some() {
+                return Err(anyhow!("Multiple derive for {} is not supported", dpath));
```
Emilgardis, line 264:
```suggestion
                return Err(anyhow!("multiple derive for {} is not supported", dpath));
```

svd-parser/src/expand.rs
```diff
@@ -0,0 +1,396 @@
...
+            }
+            f = f.derive_from(d);
+            f.derived_from = None;
+        } else {
+            return Err(anyhow!("Field {} not found", dpath));
```
Emilgardis, line 269:
```suggestion
            return Err(anyhow!("field {} not found", dpath));
```

svd-parser/src/expand.rs
```diff
@@ -0,0 +1,396 @@
...
+};
+
+#[derive(Clone, Debug, Default)]
+pub struct Index<'a> {
+    pub peripherals: HashMap<String, &'a Peripheral>,
+    pub clusters: HashMap<String, &'a Cluster>,
+    pub registers: HashMap<String, &'a Register>,
+    pub fields: HashMap<String, &'a Field>,
+    pub evs: HashMap<String, &'a EnumeratedValues>,
+}
```
Emilgardis, line 16: Is this intended to be `pub`? I'm guessing not, since no public function exposes it.

infact, all `pubs` in this module could be removed except for on the `expand` method (which should/could be `pub(crate)`

svd-parser/src/expand.rs
```diff
@@ -0,0 +1,396 @@
...
+    let peripherals = take(&mut device.peripherals);
+    for mut p in peripherals {
+        let mut path = p.name.to_string();
+        if let Some(dpath) = p.derived_from.as_ref() {
+            path = dpath.into();
+            if let Some(d) = index.get_base_peripheral(dpath) {
+                p = p.derive_from(d);
+                p.derived_from = None;
+            } else {
+                return Err(anyhow!("Peripheral {} not found", dpath));
```
Emilgardis, line 367:
```suggestion
                return Err(anyhow!("peripheral {} not found", dpath));
```

svd-parser/src/expand.rs
```diff
@@ -0,0 +1,396 @@
...
+        };
+        if let Some(d) = d {
+            if d.derived_from.is_some() {
+                return Err(anyhow!("Multiple derive for {} is not supported", dpath));
+            }
+            *ev = ev.derive_from(d);
+            ev.derived_from = None;
+        } else {
+            return Err(anyhow!(
+                "EnumeratedValues {} not found, parent field: {}, regparent: {}",
```
Emilgardis, line 343:
```suggestion
                "enumeratedValues {} not found, parent field: {}, regparent: {}",
```

svd-parser/src/expand.rs
```diff
@@ -0,0 +1,396 @@
...
+            }
+            // FIELD.EVNAME: Search in same field
+            1 => index.evs.get(&format!("{}.{}", rpath, dpath)).copied(),
+            // FULL.PATH.EVNAME:
+            2 => index.evs.get(&format!("{}.{}", regparent, dpath)).copied(),
+            _ => index.evs.get(dpath).copied(),
+        };
+        if let Some(d) = d {
+            if d.derived_from.is_some() {
+                return Err(anyhow!("Multiple derive for {} is not supported", dpath));
```
Emilgardis, line 337:
```suggestion
                return Err(anyhow!("multiple derive for {} is not supported", dpath));
```

svd-parser/src/expand.rs
```diff
@@ -0,0 +1,396 @@
...
+) -> Result<()> {
+    let cpath = if let Some(dpath) = c.derived_from.as_ref() {
+        let cpath = dpath.to_string();
+        if let Some(d) = index
+            .clusters
+            .get(dpath)
+            .or_else(|| index.clusters.get(&format!("{}.{}", path, dpath)))
+        {
+            if d.derived_from.is_some() {
+                return Err(anyhow!("Multiple derive for {} is not supported", dpath));
```
Emilgardis, line 148:
```suggestion
                return Err(anyhow!("multiple derive for {} is not supported", dpath));
```

svd-parser/src/lib.rs
```diff
@@ -40,15 +40,31 @@ pub mod types;
 #[non_exhaustive]
 pub struct Config {
     pub validate_level: ValidateLevel,
-    //pub expand_arrays: bool,
-    //pub expand_derived: bool,
+    #[cfg(feature = "expand")]
+    /// Expand arrays and derive
+    pub expand: bool,
+    /// Skip parsing `enumeratedValues` and `writeConstraint` in `Field`
```
Emilgardis, line 46:
```suggestion
    /// Skip parsing and emitting `enumeratedValues` and `writeConstraint` in [`Field`](field::FieldInfo)
```

Emilgardis (changes requested):
looks good!

I think this is in dire need of some documentation/better explanation though. Would suffice to add documentation to the `expand` field on config (declaring what it does in a more precise way) and changing the changelog entry a bit.

.gitmodules
burrbull: Ups. Instead of drop, I squashed this commit.

svd-parser/src/lib.rs
```diff
@@ -171,6 +199,9 @@ mod registerproperties;
 mod usage;
 mod writeconstraint;
 
+#[cfg(feature = "expand")]
+mod expand;
+
```
Emilgardis, line 204: maybe we should expose the function as well?

```suggestion
#[cfg(feature = "expand")]
mod expand;

#[cfg(feature = "expand")]
pub use expand::expand;
```

Emilgardis (changes requested):

svd-parser/src/lib.rs
burrbull: Thanks

Emilgardis (approved):
awesome!

bors r+

PR #202, opened 2022-05-11
nested derive

burrbull:


burrbull:
cc @Emilgardis

svd-parser/src/expand.rs
```diff
@@ -8,98 +8,152 @@ use svd_rs::{
     Device, EnumeratedValues, Field, Peripheral, Register, RegisterCluster, RegisterProperties,
 };
 
+#[derive(Clone, Debug, PartialEq, Hash, Eq)]
+pub struct RPath {
```
Emilgardis, line 12: I think it's ok to have these have a longer name

```suggestion
pub struct RegisterPath {
```
keeping it is fine too though, it may just be unclear for someone not familiar with how this works and looking here first time

svd-parser/src/expand.rs
```diff
@@ -8,98 +8,152 @@ use svd_rs::{
...
+        let name = parent.path.pop().unwrap();
+        (parent, name)
+    }
+    pub fn parent(&self) -> Self {
+        self.parent_name().0
+    }
+}
+
+#[derive(Clone, Debug, PartialEq, Hash, Eq)]
+pub struct FPath {
```
Emilgardis, line 53:
```suggestion
pub struct FieldPath {
```

svd-parser/src/expand.rs
```diff
@@ -8,98 +8,152 @@ use svd_rs::{
     Device, EnumeratedValues, Field, Peripheral, Register, RegisterCluster, RegisterProperties,
 };
 
+#[derive(Clone, Debug, PartialEq, Hash, Eq)]
+pub struct RPath {
+    peripheral: String,
```
Emilgardis, line 13:
```suggestion
    name: String,
```

svd-parser/src/expand.rs
```diff
@@ -8,98 +8,152 @@ use svd_rs::{
...
+    }
+    pub fn parent(&self) -> Self {
+        self.parent_name().0
+    }
+}
+
+#[derive(Clone, Debug, PartialEq, Hash, Eq)]
+pub struct FPath {
+    register: RPath,
+    field: String,
```
Emilgardis, line 55:
```suggestion
    ident: String,
```

or

```suggestion
    name: String,
```

Emilgardis (changes requested):

svd-parser/src/expand.rs
burrbull: Technically this is not path of register but path to register/cluster/fields. Like a directory.

burrbull:
Now it should look much cleaner

burrbull:
cc @Emilgardis

Emilgardis (approved):
mb! thought I r+ ed this earlier!

bors r+

PR #205, opened 2022-07-11
fix(svd-parser): parse `xs:noNamespaceSchemaLocation`

duskmoon314:


burrbull:
Looks like MSRV issues in svd-encoder deps:
https://github.com/bluss/indexmap/issues/234

cc @Emilgardis @therealprof What should we do?

Emilgardis:
Feel free to bump, policy for version bumps is to bump when needed

duskmoon314:
So we need to bump the rust edition to `2021` and the rust version to `1.56.0`?

Emilgardis:
updating edition is not needed, but that sounds good :)

duskmoon314:
A try to bump MSRV to 1.56.0 (2021). Hope not missing something.

svd-encoder/Cargo.toml
```diff
@@ -9,11 +9,12 @@ license = "MIT OR Apache-2.0"
 name = "svd-encoder"
 repository = "https://github.com/rust-embedded/svd"
 edition = "2018"
```
burrbull, line 11: edition

burrbull (approved):

PR #207, opened 2022-07-14
feat: add advanced encoder options `Config`

duskmoon314:
- Add `convert_case` to dependencies
- Add advanced encoder options `Config`

BREAKING CHANGE
- `Encode` trait needs to implement `encode_with_config`

### DIFF
--- a/svd-encoder/src/config.rs
+++ b/svd-encoder/src/config.rs
@@ -1,0 +1,472 @@
+use std::str::FromStr;
+
+use convert_case::{Boundary, Case, Casing};
+
+use crate::svd::BitRangeType;
+
+#[derive(Clone, Copy, Debug)]
+pub enum IdentifierFormat {
+    /// `Camel` case
+    ///
+    /// `aCamelCaseExample`
+    Camel,
+    /// `Pascal` case
+    ///
+    /// `APascalCaseExample`
+    Pascal,
+    /// `Snake` case
+    ///
+    /// `a_snake_case_example`
+    Snake,
+    /// `UpperSnake` case
+    ///
+    /// `AN_UPPER_SNAKE_CASE_EXAMPLE`
+    UpperSnake,
+}
+
+impl FromStr for IdentifierFormat {
+    type Err = ();
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
+            "Camel" => Ok(IdentifierFormat::Camel),
+            "Pascal" => Ok(IdentifierFormat::Pascal),
+            "Snake" => Ok(IdentifierFormat::Snake),
+            "UpperSnake" => Ok(IdentifierFormat::UpperSnake),
+            _ => Err(()),
+        }
+    }
+}
+
+pub fn change_case(s: &str, case: Option<IdentifierFormat>) -> String {
+    match case {
+        None => s.to_string(),
+        Some(case) => {
+            let boundary = [
+                Boundary::Underscore,
+                Boundary::Hyphen,
+                Boundary::Space,
+                Boundary::LowerUpper,
+                Boundary::UpperLower,
+                Boundary::Acronym,
+            ];
+
+            s.with_boundaries(&boundary).to_case(match case {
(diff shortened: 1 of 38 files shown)


burrbull:
Looks good in first view.
Could you also add/modify some tests in `tests` directory?

duskmoon314:
Ok, I will add some tests ASAP.

duskmoon314:
I add some simple tests to show the usage of `svd_encoder::Config`

It's good to add tests for all options but needs more work. orz

svd-encoder/src/config.rs
```diff
@@ -0,0 +1,472 @@
...
+    /// `APascalCaseExample`
+    Pascal,
+    /// `Snake` case
+    ///
+    /// `a_snake_case_example`
+    Snake,
+    /// `UpperSnake` case
+    ///
+    /// `AN_UPPER_SNAKE_CASE_EXAMPLE`
+    UpperSnake,
```
burrbull, line 24: Here http://calebmer.com/inflections/inflections/case/index.html
it is called "constant case"

duskmoon314:
- rename `UpperSnake` to `Constant`
- add tests in all files that are included in `tests/lib.rs`

burrbull (dismissed):

burrbull:
cc @Emilgardis any comments?

tests/src/lib.rs
```diff
@@ -41,6 +41,39 @@ pub fn run_test<
...
+pub fn run_test_with_config<
+    T: Parse<Error = SVDErrorAt, Object = T, Config = Config>
+        + Encode<Error = EncodeError>
+        + core::fmt::Debug
+        + PartialEq,
+>(
+    tests: &[(T, &str, &str)],
+    parser_config: Option<Config>,
+    encoder_config: Option<svd_encoder::Config>,
+) {
```
Emilgardis, line 53: I think `run_test` should use this function, so to not duplicate code

Emilgardis (changes requested):
awesome!

missing a optional serialize/deserialize for `Config`, but that's not super important and not a blocker (would just be nice to have to enable some kind of config file in higher level crate).

duskmoon314:
> missing a optional serialize/deserialize

There is no serialize/deserialize for `svd_parser::Config`, so I think it's unnecessary. `svd_encoder::Config` has many fields so it might be beneficial in use. Now I declare an `update` method to set config, and this is the way I use in the temptation of implementing the `format` feature in svdtools.

> I think `run_test` should use this function

Sure. Will fix this soon.

duskmoon314:
Change `run_test` to:

```rust
pub fn run_test<
    T: Parse<Error = SVDErrorAt, Object = T, Config = Config>
        + Encode<Error = EncodeError>
        + core::fmt::Debug
        + PartialEq,
>(
    // tests: [(Object, input string, output string)],
    tests: &[(T, &str, &str)],
    parser_config: Option<Config>,
    encoder_config: Option<svd_encoder::Config>,
)
```

Change all tests using `run_test` (Keep most of them, add several tests for config)

Emilgardis:
> There is no serialize/deserialize for `svd_parser::Config`, so I think it's unnecessary.

Indeed, just saying it's something we should maybe look into :D 

Regarding the `run_test`, I think it's worth having them as two separate functions, and `run_test` calls `runs_test_with_config`.

For that to work `run_test_with_config` should probably take a iterator instead..

However, I'm fine with this.

Emilgardis (approved):
awesome job, thanks a lot!

one minor issue/non-issue with `run_test`, but it's not a problem

duskmoon314:
> Regarding the `run_test`, I think it's worth having them as two separate functions, and `run_test` calls `runs_test_with_config`.
> 
> For that to work `run_test_with_config` should probably take a iterator instead..

Oh, you are right. That way is more simple to use. I need to think about how to do this.

burrbull (approved):

burrbull:
Lets merge this. @duskmoon314 if you have more improvements, please open new PR.
bors r+

PR #220, opened 2023-03-31
sorting on encode

burrbull:


burrbull:
cc @Emilgardis

svd-encoder/src/peripheral.rs
```diff
@@ -102,10 +104,41 @@ impl Encode for PeripheralInfo {
...
+                        sort_register_cluster(&mut refs, Some(sorting));
+                        refs.into_iter()
+                            .map(|e| e.encode_node_with_config(config))
+                            .collect()
+                    } else {
+                        v.iter()
+                            .map(|e| e.encode_node_with_config(config))
+                            .collect()
+                    }
+                };
```
Emilgardis, line 141: clippy wants to merge these, I think I agree

svd-encoder/src/config.rs
```diff
@@ -166,6 +166,44 @@ impl FromStr for FieldBitRangeFormat {
     }
 }
 
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum Sorting {
+    Offset,
+    OffsetReversed,
+    Name,
+}
+
```
Emilgardis, line 175:
```suggestion
#[derive(Clone, Copy, Debug, PartialEq, Eq)]
#[non_exhaustive]
/// Apply a kind of sorting
pub enum Sorting {
    /// Sort by offsets
    Offset,
    /// Same as [`Sorting::Offset`], but reversed
    OffsetReversed,
    /// Sort by name
    Name,
}

```

Emilgardis: It would be neat also to have a 

```
for<'a> Fn(/*adress:*/ i32, /* name: */ &'a str, ... /*more?*/) -> Ordering
```

to implement alternative sorting schemes, but not really needed

Emilgardis (changes requested):
lgtm with clippy fixed

svd-encoder/src/peripheral.rs
burrbull: Second branch is default. It should be fastest and avoid advanced allocations.

svd-encoder/src/config.rs
burrbull: Interesting suggestion but I would leave it for someone other.

burrbull: It looks like this will require Config to depends 3 generics.

Emilgardis (approved):
lgtm!

burrbull:
bors r=@Emilgardis

PR #225, opened 2023-06-04
update ci & deprecate bors

Emilgardis:
part of rust-embedded/svd2rust#724

remove all uses of actions-rs which use deprecated gha functions and is not maintained
removes clippy_check in favour of running clippy directly


Emilgardis:
I notice we don't have any caching here, should we maybe add that?

burrbull:
> I notice we don't have any caching here, should we maybe add that?

Don't see big sense. All checks are fast, rarely changes.

Needs MSRV bump.

burrbull (dismissed):

Emilgardis:
did you change the settings here also @burrbull ? The settings right now are setup for merge_grup, but that has not been implemented yet

burrbull:
> did you change the settings here also @burrbull ? The settings right now are setup for merge_group, but that has not been implemented yet

Add to this PR, please.

I also cannot add clippy restriction.

Emilgardis:
I'll add it

clippy is behind `CI` now

Emilgardis:
image

this is not a good way to do it, since if we update msrv we'd have to bump here also, we can fix it with a name or switching to a job with needs. Since this is only a few checks I'll name the job

burrbull:
> clippy is behind `build` now

I don't like this. `check` and `clippy` should be different jobs.

Emilgardis:
clippy is check though, they do exactly the same thing, just more knobs with clippy.

burrbull:
> clippy is check though, they do exactly the same thing, just more knobs with clippy.

if job fails I want to see is it build fail or just clippy.

Emilgardis:
we don't use `-D warnings` so there shouldn't be any issues with warnings from clippy

Emilgardis:
oh, we do do `#[deny(warnings)]`, that's bad :3

Emilgardis:
actually, I'd rather fix the deny first and then we can determine if we should split it, I'll let you decide how it should be afterwards

.github/workflows/ci.yml
```diff
@@ -1,62 +1,67 @@
...
-          profile: minimal
-          override: true
-      - uses: actions-rs/cargo@v1
-        with:
-          command: check
+          components: clippy
+      - run: cargo check
+        env:
+          RUSTFLAGS: -D warnings
+      - run: cargo clippy
```
Emilgardis, line 45: this should be fine imo since clippy doesn't add "error-by-default" lints just like that without reason, most new lints are added as warn or even allow

svd-parser/src/lib.rs
```diff
@@ -19,9 +19,7 @@
 //! - [SVD Schema file](https://www.keil.com/pack/doc/CMSIS/SVD/html/schema_1_2_gr.html)
 //! - [SVD file database](https://github.com/posborne/cmsis-svd/tree/master/data)
 //! - [Sample SVD file](https://www.keil.com/pack/doc/CMSIS/SVD/html/svd_Example_pg.html)
-
-#![deny(warnings)]
-
```
Emilgardis, line 24: replaced with `RUSTFLAGS=-D warnings`

burrbull (approved):

Issue #228, opened 2023-09-08
[svd_parser]: Tag that contains XML are ignored.


pellico:
svd_parser 0.14.1

If a tag contains a XML comment the tag content is ignored.
see tag `licenseText`
Below XML has been validated against latest svd.xsd.

How to reproduce:
```rust
use svd_parser;


fn main() {

    let xml=r#"<?xml version="1.0" encoding="utf-8"?>
<device schemaVersion="1.1" xmlns:xs="http://www.w3.org/2001/XMLSchema-instance" xs:noNamespaceSchemaLocation="svd.xsd">

	<name>ARM_Example</name>
	<series>ARMCM3</series>
	<version>1.2</version>
	<description>SVD Test for Rust PAC generator</description>
	<licenseText><!-- this license text will appear in header file. \n force line breaks --> Test license\n </licenseText>
	<cpu>
		<!-- details about the cpu embedded in the device -->
		<name>CM3</name>
		<revision>r1p0</revision>
		<endian>little</endian>
		<mpuPresent>true</mpuPresent>
		<fpuPresent>false</fpuPresent>
		<nvicPrioBits>3</nvicPrioBits>
		<vendorSystickConfig>false</vendorSystickConfig>
	</cpu>
	<addressUnitBits>8</addressUnitBits>
	<width>32</width>
	<size>32</size>
	<access>read-write</access>
	<resetValue>0x00000000</resetValue>
	<resetMask>0xFFFFFFFF</resetMask>
	<peripherals>
		<peripheral>
			<name>UART</name>
			<baseAddress>0x50000000</baseAddress>
			<registers>
				<register>
					<name>Reg1</name>
					<addressOffset>0x0</addressOffset>
					<resetMask>0xFFFFFFFF</resetMask>
				</register>
			</registers>
		</peripheral>
	</peripherals>
</device>
    "#;
    let device = svd_parser::parse(xml).unwrap();
    println!("{:?}",device.license_text);
    println!("{:?}",device.description);

}

```

PR #229, opened 2023-09-16
always trim whitespaces

burrbull:


burrbull:
cc @Emilgardis
I see 3 ways to fix #228 :
1. Always allow whitespaces (this PR).
2. Don't allow on `strict`:
2a. Change `ElementExt` to get `Config`. *Breaking change*, require version bump.
2b. Wrappers for `ElementExt` methods and replace them on each usage.

thejpster:
This PR only changes the CHANGELOG and nothing else?

burrbull:
Sorry

thejpster (approved):
Makes sense to me

Emilgardis:
I'm not sure I like this approach, but there shouldn't realistically be a scenario where a leading/trailing whitespace is meaningful so it's fine

burrbull:
> I'm not sure I like this approach, but there shouldn't realistically be a scenario where a leading/trailing whitespace is meaningful so it's fine

I'm going to change this in future. I just don't want to do major release for such little change.

PR #233, opened 2023-10-03
add write constraint range check

### LINKED ISSUE #232: Check writeConstraint range
The range of values for a enumeration is checked against a field size. But there is no such check for a range. It is possible to assign a range that out of bounds field.

kossnikita:
fixes #232

### DIFF
--- a/svd-rs/src/writeconstraint.rs
+++ b/svd-rs/src/writeconstraint.rs
@@ -1,3 +1,5 @@
+use super::SvdError;
+
 /// Define constraints for writing values to a field
 #[cfg_attr(
     feature = "serde",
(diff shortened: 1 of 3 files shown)


burrbull (approved):

burrbull:
rustfmt, please

svd-rs/src/writeconstraint.rs
```diff
@@ -25,3 +27,22 @@ pub struct WriteConstraintRange {
...
+    OutOfRange(u64, core::ops::Range<u64>),
+}
+
+impl WriteConstraintRange {
+    pub(crate) fn check_range(&self, range: core::ops::Range<u64>) -> Result<(), SvdError> {
+        for v in [&self.min, &self.max] {
+            if !range.contains(v) {
+                return Err(Error::OutOfRange(*v, range.clone()).into());
+            }
+        }
```
burrbull, line 45: What sense of loop here? It can be done much faster with just `if`.

burrbull (approved):

burrbull:
Possibly check also `self.min<=self.max`?

kossnikita:
> `self.min<=self.max`?

I'm not sure they can't be equal.

svd-rs/src/writeconstraint.rs
```diff
@@ -25,3 +27,28 @@ pub struct WriteConstraintRange {
...
+    #[error("Value {0} out of range {1:?}")]
+    OutOfRange(u64, core::ops::Range<u64>),
+    /// Minimum is greater than maximum.
+    #[error("Range minimum {0} is greater than maximum {1}")]
+    ReversedRange(u64, u64),
+}
+
+impl WriteConstraintRange {
+    pub(crate) fn check_range(&self, range: core::ops::Range<u64>) -> Result<(), SvdError> {
+        if &self.min > &self.max {
```
burrbull, line 44: I don't see sense in references here. Could you run `cargo clippy`, please.
Looks like I missed it in CI.

kossnikita: Yeah, you right

PR #235, opened 2023-10-23
custom names for arrays

burrbull:


burrbull:
cc @Emilgardis

svd-rs/src/array.rs
```diff
@@ -50,11 +50,51 @@ where
...
+        dim.dim_array_index
+            .as_ref()
+            .map(|dai| {
+                dai.values
+                    .iter()
+                    .find(|e| e.value.map(|v| v.to_string().as_str() == i.deref()) == Some(true))
+            })
+            .flatten()
+            .map(|n| n.description.clone())
+            .flatten()
```
Emilgardis, line 95: `Some(_).map(...).flatten()` is the same as `and_then`

svd-rs/src/array.rs
```diff
@@ -50,11 +50,51 @@ where
...
-        .map(move |i| name.replace("[%s]", &i).replace("%s", &i))
+    dim.indexes().map(move |i| {
+        dim.dim_array_index
+            .as_ref()
+            .map(|dai| {
+                dai.values
+                    .iter()
+                    .find(|e| e.value.map(|v| v.to_string().as_str() == i.deref()) == Some(true))
+            })
+            .flatten()
```
Emilgardis, line 73: `Some(_).map(...).flatten()` is the same as `and_then`

Emilgardis (changes requested):

svd-rs/src/array.rs
burrbull: Exactly. Thanks.

Emilgardis (approved):

PR #238, opened 2023-10-26
M318 and M351 checks

kossnikita:
<html>
<body>

Message Number | Type | Message Text 
-- | -- | -- 
M318 | WARNING | 'LEVEL' 'NAME' <'TAG'> is equal to \<name>
M351 | WARNING | Peripheral 'TYPE' 'NAME' is equal to Peripheral name.

</body>
</html>

Doesn't add if it's equal to name:
Peripheral:
- displayName
- description
- alternatePeripheral
- groupName
- headerStructName
Register:
- displayName
- description
- alternateRegister

I can't say that I'm happy with this method. This may be unexpected for the user.
But this is a very simple way to remove `displayName` duplicate `name` in my case. Patching this manually seems terrible.

### DIFF
--- a/svd-encoder/src/peripheral.rs
+++ b/svd-encoder/src/peripheral.rs
@@ -36,7 +36,9 @@ impl Encode for PeripheralInfo {
         ));
 
         if let Some(v) = &self.display_name {
-            elem.children.push(new_node("displayName", v.to_string()));
+            if v.ne(&self.name) {
+                elem.children.push(new_node("displayName", v.to_string()));
+            }
         }
 
         if let Some(v) = &self.version {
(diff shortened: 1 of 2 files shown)


burrbull:
> I can't say that I'm happy with this method.

Silently remove tags is very bad idea.
And this is not a place to patch.

kossnikita:
Can we do this as an option in the configuration, which can be activated by an argument in svd-tools?

burrbull:
Are those `displayName`s present in initial SVD or produced by patches?

> Can we do this as an option in the configuration, which can be activated by an argument in svd-tools?

`svdtools patch` or `svdtools convert`?

burrbull:
If we are talking about `svdtools patch` and initially present displayNames it should be formed as rule. Similar to `_clear_fields`: https://github.com/stm32-rs/stm32-rs/blob/ea6c0312dfc79e74bf3cbbcb735a3c32328dcfb2/devices/stm32g071.yaml#L6

If it is incorrectly produced by patch, it is a bug.

burrbull:
Actually I don't understand what sense of `displayName` and why it should be different from `name`.
Possibly it is something `C`-relared.

kossnikita:
> Similar to `_clear_fields`

Another undocumented feature...  😅
I will take care about it someday.

I think `displayName` should be used in debugger viewer for some cases. And it should be different because it's equal to name by default. This makes my svds much smaller. I can't imagine any others reasons. It's only warning in any case.

kossnikita:
I'm not sure I understand you. Are you proposing to introduce a new feature for clearing `displayName` similar to `_clear_fields`? Or does such a feature already exist?

burrbull:
New feature. For already present `displayName`s.

More investigation for other cases.

PR #241, opened 2023-11-10
fix indexes_as_range

burrbull:
Fixes https://github.com/rust-embedded/svdtools/pull/178#issuecomment-1801433808

### DIFF
--- a/svd-rs/src/dimelement.rs
+++ b/svd-rs/src/dimelement.rs
@@ -170,6 +170,8 @@ impl DimElement {
     pub fn indexes_as_range(&self) -> Option<RangeInclusive<u32>> {
         let mut integers = Vec::with_capacity(self.dim as usize);
         for idx in self.indexes() {
+            // XXX: indexes that begin with leading zero are not compatible with range (`0-x`) syntax in serialization
+            // see https://github.com/rust-embedded/svdtools/pull/178#issuecomment-1801433808
             let val = idx.parse::<u32>().ok()?;
             if val.to_string() != idx {
                 return None;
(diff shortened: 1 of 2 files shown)


burrbull:
cc @Emilgardis

svd-rs/src/dimelement.rs
```diff
@@ -170,7 +170,11 @@ impl DimElement {
     pub fn indexes_as_range(&self) -> Option<RangeInclusive<u32>> {
         let mut integers = Vec::with_capacity(self.dim as usize);
         for idx in self.indexes() {
-            integers.push(idx.parse::<u32>().ok()?);
+            let val = idx.parse::<u32>().ok()?;
+            if val.to_string() != idx {
+                return None;
+            }
```
Emilgardis, line 176: what exactly does this fix?

that `01` != `1`?

Emilgardis: to note, I think it's the wrong way to fix the issue but maybe it's not

burrbull: > what exactly does this fix?
> 
> that `01` != `1`?

Yes

Emilgardis: can we add a comment here to explain why we need the name to be an exact match with the integer representation

burrbull: could you add comment yourself, please? I have issues with internet, maybe several days

Emilgardis (approved):
feel free to merge if the comment looks good

Emilgardis:
ci stuck on waiting for runner 
image

PR #246, opened 2023-11-22
release svd-rs-0.14.5

burrbull:


svd-rs/CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+## [v0.14.5] - 2023-11-22
+
 - `default_value` for `EnumeratedValues`
 - move `EnumeratedValue` check to `strict` mode and fix it
```
Emilgardis, line 13: hmm reading this, is this actually true?

Emilgardis (changes requested):

svd-rs/CHANGELOG.md
burrbull: what you mean?

Emilgardis: it's too broad, it should say something like `moved isDefault and value check on enumerated values to strict mode`

Emilgardis (approved):
:shipit:

PR #253, opened 2023-11-29
release svd-rs 0.14.6

burrbull:


Emilgardis (approved):
I thought going from &mut self to &self was a breaking change, but it's not!

LGTM!

PR #256, opened 2024-01-03
release 0.14.7

burrbull:
`regex` 1.10 requires rust 1.65. So I've fixed it to `1.9`.
cc @Emilgardis

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,7 +29,7 @@ jobs:
           - rust: nightly
             experimental: true
           # MSRV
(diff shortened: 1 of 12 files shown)


Emilgardis (approved):

bugadani:
This is a bit drastic and seems to cause problems with projects that don't fix regex. Wouldn't a normal-ish `regex = "1.9"` have sufficed?

burrbull:
Yeah. Looks like it was bad idea. We should just up MSRV.

Dirbaio:
this broke chiptool too. :)

PR #270, opened 2024-08-20
Add `Exception` element to RISC-V builder

romancardenas:
As part of my work in [this PR](https://github.com/rust-embedded/riscv/pull/223), I noticed that I forgot a mechanism to define custom exception sources. This PR should fix this.

### DIFF
--- a/svd-encoder/CHANGELOG.md
+++ b/svd-encoder/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+- Adapt the `riscv` element to handle `riscv::Exception`.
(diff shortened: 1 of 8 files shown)


burrbull:
Is `Exception` structure same as `Interrupt`?
We shoud reuse it instead of creating a copy.

romancardenas:
(Sorry, I used the sync button in GitHub)

I used copies so the tag ID is `exception` instead of `interrupt`. In RISC-V literature, it is common to see a clear separation between exceptions (synchronous events associated with an instruction) and interrupts (external event that occurs asynchronously to the current running thread). In any case, I don't have a strong opinion on this, so if you prefer, I can use `interrupt` elements instead of the current `exception` copy.

burrbull (approved):

PR #272, opened 2024-08-22
clippy fixes

burrbull:


svd-parser/src/expand.rs
```diff
@@ -63,11 +61,11 @@ impl PartialEq<str> for BlockPath {
         }
         let mut parts = other.split('.');
         if let Some(part1) = parts.next() {
-            if self.peripheral != part1 {
+            if self.peripheral.deref() != part1 {
```
jannic, line 64: What clippy message is fixed by that change?

PR #274, opened 2024-09-29
adding children to cluster modify

cosmikwolf:
when using cluster modify, it only adds the incoming children, and deletes the existing children

this just extends the existing children array with the incoming children


cosmikwolf:
related to https://github.com/rust-embedded/svdtools/issues/245 

@burrbull implemented this manually in svdtool, but I think it might be more appropriate to fix it here? not sure what this would break for existing implementations though...

burrbull:
> when using cluster modify, it only adds the incoming children, and deletes the existing children

It is how it works.

Please, use svdtools `_add` if you want to add something, not a `_modify`.

PR #277, opened 2024-11-14
Make access parsing case-insensitive.

Dirbaio:
SVDs from Nordic are using `read-writeonce` instead of `read-writeOnce`.


burrbull:
This looks not good for me as CMSIS-SVD does not allow this.
If you really want this, it shoud be done on `svd-parser` level when `ValidateLevel != Strict`.

PR #283, opened 2025-01-22
Do not derive DimElement

burrbull:
cc @Emilgardis 

Now I think it is incorrect behavior. So this is bug fix.


burrbull:
This one too, please.

Emilgardis (approved):

PR #286, opened 2025-02-10
Partially revert #283

burrbull:
@robamu You issue may be related to #283. Try to depend on this change.

### DIFF
--- a/svd-rs/CHANGELOG.md
+++ b/svd-rs/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+- Partially revert #283
+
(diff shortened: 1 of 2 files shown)


robamu:
Thanks, that appears to fix the issue

burrbull:
cc @Emilgardis look like we should merge this and yank 0.4.10. Check of name should cover both derive cases.

Emilgardis:
Sorry, I did kinda notice this when reviewing but didn't question it!

Emilgardis (approved):

burrbull:
@robamu
As SVD specification does not say anything about how "derive" from array should work when no dim is specified I'd recommend to specify it always.

robamu:
I suppose I could patch it. Thanks for the quick fixes!

PR #289, opened 2025-02-15
Add missing `DeriveFrom` entry for `address_block`

### LINKED ISSUE #288: `DeriveFrom` for `PeripheralInfo` missing address blocks
> Specify an address range uniquely mapped to this peripheral. A peripheral must have at least one address block, but can allocate multiple distinct address ranges. If a peripheral is derived from another peripheral, the addressBlock is not mandatory.

According to the CMSIS spec a peripheral _must_ have an address block, only if the peripheral derives from another can it be omitted.

In the `DeriveFrom` for `PeripheralInfo` the address block is not copied over, making the peripheral without an address block.

https://github.com/rust-embedded/svd/blob/3019dbc8a52f6365b0d83d6342ac836543842058/svd-rs/src/derive_from.rs#L40-L64

My use case is using the `svd_parser` crate with the `expand` feature, meaning that after I expand my device I am without the `derivedFom` entry, making it rather annoying to have to resolve the address blocks before expanding.

emesare:
Fixes #288

I didn't look much into it, but i'm pretty sure we only copy the address blocks if the entry is absent in the derived info.

### DIFF
--- a/svd-rs/CHANGELOG.md
+++ b/svd-rs/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+- Add missing entry for `address_block` in `DeriveFrom` impl
+
 ## [v0.14.11] - 2025-02-10
 
 - Partially revert #283. Check name on "%s" when derive array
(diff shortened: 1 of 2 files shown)


burrbull:
Format code and update https://github.com/rust-embedded/svd/blob/master/svd-rs/Cargo.toml please.

emesare:
Apologies for forgetting to format! What would you like me to do in the `Cargo.toml`?

Thanks for the quick response 😄

burrbull:
Wrong link. I meant changelog https://github.com/rust-embedded/svd/blob/master/svd-rs/CHANGELOG.md

emesare:
Done! :)

burrbull (approved):

PR #293, opened 2025-09-14
Handle overrides in element group derivation

scootermon:
I have an SVD file with two instances of a peripheral. The second peripheral contains the same registers, but instead of 128 it only has 64. The SVD file uses `derivedFrom` in the second peripheral to inherit these registers but overrides the `<dim>` element. To my surprise, `svd2rust` generated arrays with the same number of elements for both instances.
Well, after following the trail I ended up here and the reason wasn't hard to find.

I saw #283 and the follow up #286, but there's little to no reasoning attached to these PRs.

Looking at the spec [here](https://open-cmsis-pack.github.io/svd-spec/main/elem_registers.html#elem_register) I really don't see a reason why the fields from `dimElementGroup` shouldn't also be overridden.
The `derivedFrom` description reads

> Specify the register name from which to inherit data. Elements specified subsequently override inherited values. 

I can't see a statement about the `dimElementGroup` elements somehow being exempt.

### DIFF
--- a/svd-rs/src/derive_from.rs
+++ b/svd-rs/src/derive_from.rs
@@ -1,6 +1,6 @@
 //! Implementations of DeriveFrom, setting non-explicit fields.
 use crate::{
-    ClusterInfo, EnumeratedValues, FieldInfo, MaybeArray, PeripheralInfo, RegisterInfo,
+    ClusterInfo, DimElement, EnumeratedValues, FieldInfo, MaybeArray, PeripheralInfo, RegisterInfo,
     RegisterProperties,
 };
 
@@ -110,6 +110,27 @@ impl DeriveFrom for FieldInfo {
     }
 }
 
+impl DeriveFrom for DimElement {
+    fn derive_from(&self, other: &Self) -> Self {
+        let Self {
+            dim,              // mandatory
+            dim_increment: _, // mandatory
+            dim_index,
+            dim_name,
+            dim_array_index,
+        } = other;
+
+        let mut derived = self.clone();
+        derived.dim_name = derived.dim_name.or_else(|| dim_name.clone());
+        if derived.dim == *dim {
+            // Only derive dimIndex and dimArrayIndex if the dimensions are the same.
+            derived.dim_index = derived.dim_index.or_else(|| dim_index.clone());
+            derived.dim_array_index = derived.dim_array_index.or_else(|| dim_array_index.clone());
+        }
+        derived
+    }
+}
+
 impl<T> DeriveFrom for MaybeArray<T>
 where
     T: DeriveFrom + crate::Name,
@@ -128,8 +149,11 @@ where
                     Self::Single(info.derive_from(other_info))
                 }
             }
-            (Self::Array(info, dim), Self::Single(other_info) | Self::Array(other_info, _)) => {
+            (Self::Array(info, dim), Self::Single(other_info)) => {
                 Self::Array(info.derive_from(other_info), dim.clone())
+            }
+            (Self::Array(info, dim), Self::Array(other_info, other_dim)) => {
+                Self::Array(info.derive_from(other_info), dim.derive_from(other_dim))
             }
         }
     }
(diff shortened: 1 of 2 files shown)


burrbull:
> I have an SVD file with two instances of a peripheral.

@scootermon show this file, please. Or part of the file.

svd-rs/src/derive_from.rs
```diff
@@ -110,6 +116,24 @@ impl DeriveFrom for FieldInfo {
...
+            dim_increment: _, // mandatory
+            dim_index,
+            dim_name,
+            dim_array_index,
+        } = other;
+
+        let mut derived = self.clone();
+        derived.dim_index = derived.dim_index.or_else(|| dim_index.clone());
+        derived.dim_name = derived.dim_name.or_else(|| dim_name.clone());
+        derived.dim_array_index = derived.dim_array_index.or_else(|| dim_array_index.clone());
```
burrbull:
```suggestion
        derived.dim_index = derived.dim_index.or_else(|| {
            dim_index
                .as_ref()
                .and_then(|di| (*dim == derived.dim).then(|| di.clone()))
        });
        derived.dim_name = derived.dim_name.or_else(|| dim_name.clone());
        derived.dim_array_index = derived.dim_array_index.or_else(|| {
            dim_array_index
                .as_ref()
                .and_then(|dai| (*dim == derived.dim).then(|| dai.clone()))
        });
```
We should derive `dim_index` only when same size.

scootermon:
> > I have an SVD file with two instances of a peripheral.
> 
> @scootermon show this file, please. Or part of the file.

@burrbull, unfortunately, I cannot share the file or significant portions from it. Here's a reduced version just to illustrate the pattern:

```xml
<!-- CAN0 -->
<register>
  <name>eRAM[%s]</name>
  <description>Embedded RAM</description>
  <size>32</size>
  <access>read-write</access>
  <dim>128</dim>
  <dimIncrement>4</dimIncrement>
  <!-- [Other elements omitted] -->
</register>

<!-- CAN1 (this one is complete!) -->
<register derivedFrom="CAN0.eRAM[%s]">
  <name>eRAM[%s]</name>
  <description>Embedded RAM</description>
  <dim>64</dim>
  <dimIncrement>4</dimIncrement>
</register>
```

The CAN1 peripheral only has half the amount of "RAM" registers, but the fields are otherwise identical.

Emilgardis:
Really strange that CI hasn't run.
Might be because this pr is based on [inomotech-foss:master](https://github.com/inomotech-foss/svd/tree/master) instead of a non-main branch

I agree with being hesitant to this change. Does the svd pass through svdconv?

scootermon:
> 2. You have not answer on my previous suggestion.

Apologies, @burrbull. I first wanted to settle whether there's even a chance for this PR to get merged before going into the details.

> 1. I don't understand why old code does not work for your example. Unless there's something important hidden under `<!-- [Other elements omitted] -->`

No, there's nothing important hidden there, it's just a lot more registers that aren't relevant to the issue.
Running the current `svd2rust` version on the SVD file from my gist (https://gist.github.com/scootermon/68828d5bc4f19b06dfb3a0b1e473d155#file-test-svd) generates incorrect code.
It generates both `struct CAN0` and `struct CAN1` with `eram: [Eram; 128]`, but the can1 instance doesn't have 128 instances of the `ERAM` register, only 64!

burrbull:
@scootermon What status of this?

scootermon:
> @scootermon What status of this?

If memory serves me right I believe I was waiting for a re-review from your end?

burrbull:
What about my suggestion to check dim size?

scootermon:
> What about my suggestion to check dim size?

Are you talking about something different than this?
https://github.com/rust-embedded/svd/compare/44af50ab60038de4005ed5ac52e62b3cb962f07b..136b46f65abffbe7856830c795eca794cff9f7d1

If yes, then I must've misunderstood...

burrbull:
> > What about my suggestion to check dim size?
> 
> Are you talking about something different than this? https://github.com/rust-embedded/svd/compare/44af50ab60038de4005ed5ac52e62b3cb962f07b..136b46f65abffbe7856830c795eca794cff9f7d1
> 
> If yes, then I must've misunderstood...

Sorry. Missed this.

burrbull (dismissed):

burrbull:
Please, rebase.

scootermon:
> Please, rebase.

Done. For reference, the only conflict was in the CHANGELOG file. Everything else is still the same.

burrbull (approved):

PR #294, opened 2025-09-15
Make the URL for the submodule valid

glaeqen:
CMSIS-SVD repository was moved plus git:// makes the git command hang.

[Recent
change](https://github.com/NixOS/nixpkgs/commit/5cd869caa4ea1bf1bde71731ecc96b24e1bdcc2d) in the Rust platform in `NixOS/nixpkgs` prevents a user from packaging a Rust program using crates from this repository to be used as dependencies as it forces their submodules to be checked out, regardless of the `submodule.*.update` setting.

### DIFF
--- a/svd-parser/src/elementext.rs
+++ b/svd-parser/src/elementext.rs
@@ -8,7 +8,7 @@ use super::{Parse, SVDError, SVDErrorAt};
 
 /// Defines extensions for implementation over roxmltree::Node
 pub trait ElementExt {
-    fn get_child<K>(&self, k: K) -> Option<Node>
+    fn get_child<K>(&self, k: K) -> Option<Node<'_, '_>>
     where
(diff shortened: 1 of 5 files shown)


adamgreig (dismissed):

glaeqen:
I decided to do it all at once in this PR because I realized that URL I switched to leads to a wrong repository anyway (should have been `cmsis-svd-data`, not `cmsis-svd`) and I figured it's maybe worth a shot to make this script useful, maybe even enable it in the CI. But:

Script works and generates Rust files in a way it was, I think, intended. But I'm not sure how they should be used, `svd-tests` does not try to include the generated root `cmsis.rs` file to make `cargo test` operational, not sure how these tests were supposed to be run to begin with.

Maybe it's enough for this PR.

PR #301, opened 2026-03-16
Change parsing of the CPU block to have optional mpuPresent and

### LINKED ISSUE #300: [svd parser] CPU block incorrectly requires mpuPresent and fpuPresent elements
I noticed a discrepancy between the documentation and the actual SVD schema for these 2 elements; see Open-CMSIS-Pack/svd-spec#32

I've send a PR to update the documentation to conform to the SVD schema. This change also needs a loosening in the svd-parser here... Default to `false` if these elements are not present.

xoriath:
fpuPresent elements

Per the SVD schema; these 2 elements are actually optional. However the documentation page for the schema incorrectly claims them to be required.

Modify the parser to allow these to be optional when parsing.

Fixes #300

### DIFF
--- a/svd-parser/src/cpu.rs
+++ b/svd-parser/src/cpu.rs
@@ -20,12 +20,22 @@ impl Parse for Cpu {
             _ => false,
         };
 
+        let has_mpu_present = match tree.get_child_bool("mpuPresent") {
+            Ok(v) => v,
+            _ => false,
+        };
+
+        let has_fpu_present = match tree.get_child_bool("fpuPresent") {
+            Ok(v) => v,
+            _ => false,
+        };
+
         Cpu::builder()
             .name(tree.get_child_text("name")?)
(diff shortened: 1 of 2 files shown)


svd-rs/src/cpu.rs
```diff
@@ -159,13 +159,13 @@ impl CpuBuilder {
         self
     }
     /// Set the mpu_present of the cpu.
-    pub fn mpu_present(mut self, value: bool) -> Self {
-        self.mpu_present = Some(value);
+    pub fn mpu_present(mut self, value: Option<bool>) -> Self {
```
burrbull: This looks like a breaking change. Could we use `impl Into<Option<bool>>` or something to get around the problem?

burrbull (approved):

xoriath:
What happened to the merge checks ?

xoriath:
Thanks @burrbull !

